PR #2, opened 2021-10-31
Add subtitle editor

bigfoodK:
image

### Features

- Subtitle Preview
- Edit subtitle properties
  - Font `text` `serif` `size` `color`
  - Background `color`
  - Border `color` `width`
  - Drop shadow `color` `x` `y`

### DIFF
--- a/namui/src/ColorUtil.ts
+++ b/namui/src/ColorUtil.ts
@@ -32,6 +32,34 @@ export namespace ColorUtil {
   ): CanvasKit.Color {
     return Color0255(grayscale, grayscale, grayscale, a);
   }
+  export function ColorHSL01(h: number, s: number, l: number, a?: number) {
+    const hue = (h * 360) % 360;
+    const hueStage = hue / 60;
+    const primaryChroma = (1 - Math.abs(2 * l - 1)) * s;
+    const secondaryChroma = primaryChroma * (1 - Math.abs((hueStage % 2) - 1));
+    const [baseR, baseG, baseB] =
+      hueStage < 1
+        ? [primaryChroma, secondaryChroma, 0]
+        : hueStage < 2
+        ? [secondaryChroma, primaryChroma, 0]
+        : hueStage < 3
+        ? [0, primaryChroma, secondaryChroma]
+        : hueStage < 4
+        ? [0, secondaryChroma, primaryChroma]
+        : hueStage < 5
+        ? [secondaryChroma, 0, primaryChroma]
+        : hueStage < 6
+        ? [primaryChroma, 0, secondaryChroma]
+        : [0, 0, 0];
+    const lightnessFactor = l - primaryChroma / 2;
+    return Color01(
+      baseR + lightnessFactor,
+      baseG + lightnessFactor,
+      baseB + lightnessFactor,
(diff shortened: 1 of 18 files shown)


namui/src/ColorUtil.ts
```diff
@@ -32,6 +32,34 @@ export namespace ColorUtil {
...
+        : [0, 0, 0];
+    const lightnessFactor = l - primaryChroma / 2;
+    return Color01(
+      baseR + lightnessFactor,
+      baseG + lightnessFactor,
+      baseB + lightnessFactor,
+      a,
+    );
+  }
+
```
bigfoodK, line 62: ### Notice for Engine Developers
ColorUitl supports hsl color

namui/src/getGcCanvasKitPackage.ts
```diff
@@ -45,3 +52,14 @@ function makeItGc<T extends EmbindObject<any>>(
...
+function makeItGcAlt<T extends EmbindObject<any>, P extends any[]>(
+  constructor: (...params: P) => T,
+  garbages: EmbindObject<any>[],
+) {
+  return function (...params: P): T {
+    const shader = constructor(...params);
+    garbages.push(shader);
+    return shader;
+  };
+}
```
bigfoodK, line 65: ### Notice for Engine Developers
Create a new GCMaker function for `CanvasKit.Shader.MakeLinearGradient`

### Why new function
`makeItGc()` works with constructor like `new ()`. But `CanvasKit.Shader.MakeLinearGradient` is not like that.

### What I want to ask the reviewer

1. Is it the right way?
2. Is it the right name?

bigfoodK:
`border width`, `drop shadow position` can be edited in pixel units.
Would it be better to use another method? like

1. Edit in enum like font size
2. Depending on the font size

namui/src/ColorUtil.ts
namse: GOOD TO MERGE

namse:
> `border width`, `drop shadow position` can be edited in pixel units. Would it be better to use another method? like
> 
> 1. Edit in enum like font size
> 2. Depending on the font size

I think that values should follow the enum value of font size.

namui/src/getGcCanvasKitPackage.ts
```diff
@@ -45,3 +52,14 @@ function makeItGc<T extends EmbindObject<any>>(
 
   return gcConstructor as any;
 }
+
+function makeItGcAlt<T extends EmbindObject<any>, P extends any[]>(
```
namse, line 56: I believe this would be merge with above function(`makeItGc`) because params always passed in `makeItGc` too. Could you check it and update it?

namse (changes requested):

bigfoodK:
> > `border width`, `drop shadow position` can be edited in pixel units. Would it be better to use another method? like
> > 
> > 1. Edit in enum like font size
> > 2. Depending on the font size
> 
> I think that values should follow the enum value of font size.

https://github.com/NamseEnt/namseent/pull/2/commits/9ababab39c6a12f42060bece54dd18f78bed76e5
Used `method 2`

PR #22, opened 2021-11-10
Generate random color from string

### LINKED ISSUE #19: [TODO] Random color generator with clip's id
# Entry Point
https://github.com/NamseEnt/namseent/blob/feature/scene-editor/luda/editor/client/src/timeline/clip/ClipComponent.ts

# What result I am expecting
- Set clip's color randomly
- Use clip's id as random seed

# Why do you need it
Now, the clips all have the same color.
So it is not easy to distinguish the clips.
Especially, subtitle clips require this function because the head and tail are separated and other clips can be inserted in the middle.

bigfoodK:
It uses functions similar to, but not equal to, crc32 hash functions.

`ColorUtil.getRandomColorFromString(seed: string, withAlpha?: boolean)`
Color range is rgba(0-255, 0-255, 0-255, 0-1)
If `withAlpha` option activated, color's alpha also will be random. If not, alpha will be `1`.

It closes #19

### DIFF
--- a/luda/editor/client/src/timeline/clip/SubtitleClipComponent.ts
+++ b/luda/editor/client/src/timeline/clip/SubtitleClipComponent.ts
@@ -48,10 +48,7 @@ export const SubtitleClipComponent: Render<
     x: width - componentWidth,
     y: height - componentHeight,
   };
-  const color =
-    timelineState.selectedClip?.id === clip.id
-      ? ColorUtil.Green
-      : ColorUtil.getRandomColorFromString(clip.id);
+  const color = ColorUtil.getRandomColorFromString(clip.id);
   const brighterColor = ColorUtil.brighterColor01(color, 0.2);
 
   const strokePath = new CanvasKit.Path()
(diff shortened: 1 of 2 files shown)


namui/src/ColorUtil.ts
```diff
@@ -66,6 +66,27 @@ export namespace ColorUtil {
...
+    withAlpha: boolean = false,
+  ) {
+    const hash = seed.split("").reduce((hash, character) => {
+      hash ^= character.charCodeAt(0);
+      for (let i = 0; i < 8; i++) {
+        hash = (hash >>> 1) ^ (0xedb88320 & -(hash & 1));
+      }
+      return hash;
+    }, 0xffffffff);
+
```
bigfoodK, line 80: It is similar to the crc32 hash function but does not work same.

bigfoodK:
image
**COLORFUL**

luda/editor/client/src/timeline/clip/SubtitleClipComponent.ts
```diff
@@ -51,7 +51,7 @@ export const SubtitleClipComponent: Render<
   const color =
     timelineState.selectedClip?.id === clip.id
       ? ColorUtil.Green
```
namse, line 53: I think we don't have to change the color of selected one. Would you highlight border only?

namse: Because may there are other green clips.

namse (changes requested):

namse:
image

PR #24, opened 2021-11-11
Remove subtitle clip from context menu

### LINKED ISSUE #23: [TODO] Remove subtitle clip from context menu
# Entry Point
https://github.com/NamseEnt/namseent/blob/feature/scene-editor/luda/editor/client/src/timeline/renderSubtitleTrackBody.ts

# What result I am expecting
- On right mouse button up on subtitle clip, Open context menu about clip.
- In context menu, Subtitle clip should be able to be removed.

# Why do this?
- We should remove unusing subtitle clip

bigfoodK:
If right button up on subtitle clip, It opens clip context menu, Not track context menu

closes #23


namse:
GJGJ

PR #37, opened 2021-11-25
Fix saver to update lastValueString on save

bigfoodK:
~~I found an error, so I'll reverts it.~~

Fix saver to update lastValueString on save

### DIFF
--- a/luda/editor/client/src/saver/saver.ts
+++ b/luda/editor/client/src/saver/saver.ts
@@ -49,6 +49,7 @@ class Saver implements ISaver {
       try {
         await fileSystem.write(key, stringifiedValue);
         this.saveStates.set(key, AutoSaveState.saved);
+        this.lastValueStrings.set(key, stringifiedValue);
         return;
       } catch (error) {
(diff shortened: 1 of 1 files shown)


luda/editor/client/src/saver/saver.ts
```diff
@@ -1,71 +1,41 @@
...
-    this.saveStates.set(key, AutoSaveState.saving);
 
     const stringifiedValue = this.stringify(value);
 
-    const maxRetrial = 5;
-    const retrialDelay = 1000;
-    for (let trial = 0; trial < maxRetrial; trial++) {
-      try {
-        await fileSystem.write(key, stringifiedValue);
-        this.saveStates.set(key, AutoSaveState.saved);
```
bigfoodK, line 51: Maybe we should set `lastValueStrings` here.
With `this.lastValueStrings.set(key, stringifiedValue);`
Otherwise, it will save indefinitely.

bigfoodK:
Rename and reopen to fix error

PR #41, opened 2021-11-29
Support cli to create project

namse:
I created cli to create project using cli.


common/cli/src/bin.ts
```diff
@@ -0,0 +1,13 @@
+import { Command } from "commander";
+import { createProject } from "./createProject";
+
+const program = new Command();
+
+program
+  .command("create <projectName>")
+  .description("create project")
```
namse, line 8: Because of it, we don't need README because it log how to use it itself.

common/tsconfig/package.json
```diff
@@ -1,7 +1,7 @@
 {
-    "name": "common-tsconfig",
-    "version": "4.4.3",
-    "dependencies": {
-        "namseent": "workspace:^1.0.0"
```
namse, line 5: I removed `namseent` from `common-tsconfig` to put `cli` to root `package.json`.

common/cli/src/bin.ts
namse: image

common/tsconfig/tsconfig.json
```diff
@@ -1,3 +1,102 @@
 {
-  "extends": "namseent/tsconfig.json"
+  "compilerOptions": {
```
namse, line 2: I tried to make it `../../tsconfig.json` but it make a circular dependency because directory becomes in node_modules.

PR #55, opened 2022-01-05
Fetch sequence from server

bigfoodK:
# What I did

## Define basic strategy of fetching sequence
It seems to be very event dependent

### Some Editor events added
- SequenceLoadEvent
- SequenceLoadStateUpdateEvent
- PageChangeEvent

### One rpc added
- `read_file` returns `vec<u8>`

### Procedure
1. Fire `SequenceLoadEvent` (by the user).
2. `Editor` starts fetching when editor got `SequenceLoadEvent`.
  2.1 Fire `SequenceLoadStateUpdateEvent` to set loading state to `Loading`. And request `read_file` rpc.
  2.2 When `read_file` rpc returns response.
    -  Fire `SequenceLoadStateUpdateEvent` to set loading state to `Loaded` If response is ok.
    -  Fire `SequenceLoadStateUpdateEvent` to set loading state to `Failed` If response is err.
3. Fire `PageChangeEvent` with loaded sequence to change page and clone sequence from loaded sequence to timeline's sequence


luda-editor/luda-editor-client/src/editor/mod.rs
```diff
@@ -24,10 +28,13 @@ struct Editor {
     job: Option<Job>,
     timeline: Timeline,
     clip_editor: ClipEditor,
+    sequence_list_view: SequenceListView,
     playback_time: chrono::Duration,
     socket: luda_editor_rpc::Socket,
     screen_wh: namui::Wh<f32>,
     image_filename_objects: Vec<ImageFilenameObject>,
+    sequence_load_state_map: SequenceLoadStateMap,
```
bigfoodK, line 36: Editor have `sequence_load_state_map`: `HashMap<String: path, SequenceLoadState>`;

luda-editor/luda-editor-client/src/editor/mod.rs
```diff
@@ -112,6 +119,84 @@ impl namui::Entity for Editor {
...
+                        }
+                    });
+                }
+                EditorEvent::SequenceLoadStateUpdateEvent { path, state } => match state {
+                    Some(state) => {
+                        if let Some(old_state) = self.sequence_load_state_map.get(path) {
+                            if old_state.started_at > state.started_at {
+                                return;
+                            }
+                        }
```
bigfoodK, line 181: If `SequenceLoadStateUpdateEvent` is old event. Ignore update

luda-editor/luda-editor-client/src/editor/types/mod.rs
```diff
@@ -127,19 +133,20 @@ impl SubtitlePlayDurationMeasurer {
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Subtitle {
     pub id: String,
     pub language_text_map: HashMap<Language, String>,
 }
 
+#[derive(Serialize, Deserialize, Clone)]
 pub struct Sequence {
     pub tracks: Vec<Track>,
 }
```
bigfoodK, line 152: Sequence can be `Serialize`, `Deserialize` and `Clone`

luda-editor/luda-editor-client/src/editor/mod.rs
namse: `Editor` itself has the context of single sequence editing, so I think it's not a good idea to put multiple sequence's loading information or page information inside of `Editor`. 
Would you like to make a upper object to handle that?

luda-editor/luda-editor-client/src/editor/mod.rs
namse: To achieve what you want, I guess sequence_load_state_map should be updated before sending rpc at here https://github.com/NamseEnt/namseent/pull/55/files#diff-01d4ae1db8edb49f41fba1d5f096b9f6ee34f66399f1eda95feab765eaaca687R147.
Let's think about this case,

Request A and Request B,
Request A started at 00:00
Request B started at 00:01
Request B finished at 00:02, nothing in map, so it updates map
Request A finished at 00:03, B is older than A, so it updates map.

It depends on the definition what is the 'old event'. Is it old if event fired later or rpc started later? If rpc-later is older, please update this logic like I said.

luda-editor/luda-editor-client/src/editor/mod.rs
```diff
@@ -112,6 +119,84 @@ impl namui::Entity for Editor {
...
+
+                        self.sequence_load_state_map
+                            .insert(path.clone(), (*state).clone());
+                    }
+                    None => {
```
namse, line 186: What's the meaning of None for `Option<SequenceLoadState>`?

luda-editor/luda-editor-client/src/editor/mod.rs
bigfoodK: Aaah. you mean `Main` in `main.rs` right?
I was mistaken for `Editor` to be top object.
I will fix it

luda-editor/luda-editor-client/src/editor/mod.rs
bigfoodK: Actually, status is updated before sending rpc.
And started_at means when Editor got `SequenceLoadEvent`. Regardless of when request was finished.

When Editor got `SequenceLoadEvent`. It fires `SequenceLoadStateUpdateEvent` immediately to set state to `Loading`.
And LoadState's started_at means this moment (When Editor got `SequenceLoadEvent`).
https://github.com/NamseEnt/namseent/pull/55/files#diff-01d4ae1db8edb49f41fba1d5f096b9f6ee34f66399f1eda95feab765eaaca687R123-R130

And then send rpc, update result
https://github.com/NamseEnt/namseent/pull/55/files#diff-01d4ae1db8edb49f41fba1d5f096b9f6ee34f66399f1eda95feab765eaaca687R131-R174

So your example works like as follows
Request A and Request B,
1. Request A started at 00:00. **Update state to `Loading:started at 00:00`**
2. Request B started at 00:01. **Update state to `Loading:started at 00:01`** because Request B is newer
3. Request B finished at 00:02, There is `Loading:started at 00:01`. Request B is not older than `Loading:started at 00:01`. Update to `Loaded:started at 00:01`.
4. Request A finished at 00:03, B is **newer** than A, so update(`Loaded:started at 00:00`) **will be ignored**.

luda-editor/luda-editor-client/src/editor/mod.rs
bigfoodK: It means `clearing LoadState`.
`clearing LoadState` will probably be necessary to make `sequence list view`.
In this map, key is path of sequence file. So if the path changes with action like renaming. Old path's state should be cleared.

luda-editor/luda-editor-client/src/editor/mod.rs
namse: I don't think main.rs is right place. New concept like App needed

bigfoodK:
WIP
# Changes

## App structure changed
Files gonna be
<details><summary>open</summary>
<p>
```
src/
├── app
│   ├── app.rs
│   ├── editor
│   │   ├── clip_editor
│   │   │   ├── camera_clip_editor
│   │   │   │   ├── image_browser
│   │   │   │   │   ├── back_button.rs
│   │   │   │   │   ├── browser_item.rs
│   │   │   │   │   ├── mod.rs
│   │   │   │   │   └── scroll.rs
│   │   │   │   ├── mod.rs
│   │   │   │   ├── preview.rs
│   │   │   │   └── wysiwyg_editor
│   │   │   │       ├── cropper.rs
│   │   │   │       ├── mod.rs
│   │   │   │       └── resizer.rs
│   │   │   └── mod.rs
│   │   ├── events.rs
│   │   ├── job
│   │   │   ├── mod.rs
│   │   │   ├── move_camera_clip.rs
│   │   │   ├── move_subtitle_clip.rs
│   │   │   ├── wysiwyg_crop_image.rs
│   │   │   ├── wysiwyg_move_image.rs
│   │   │   └── wysiwyg_resize_image.rs
│   │   ├── mod.rs
│   │   ├── timeline
│   │   │   ├── mod.rs
│   │   │   ├── playback_time_view.rs
│   │   │   ├── time_ruler
│   │   │   │   └── mod.rs
│   │   │   ├── timeline_body
│   │   │   │   ├── mod.rs
│   │   │   │   └── track_body
│   │   │   │       ├── camera_track_body
│   │   │   │       │   ├── camera_clip_body.rs
│   │   │   │       │   └── mod.rs
│   │   │   │       ├── mod.rs
│   │   │   │       └── subtitle_track_body
│   │   │   │           ├── mod.rs
│   │   │   │           └── subtitle_clip_body.rs
│   │   │   ├── timeline_header.rs
│   │   │   └── track_header.rs
│   │   └── types
│   │       └── mod.rs
│   ├── events.rs
│   ├── main.rs
│   ├── mod.rs
│   ├── router.rs
│   ├── sequence_list
│   │   ├── button_background.rs
│   │   ├── button_text.rs
│   │   ├── events.rs
│   │   ├── load_button.rs
│   │   ├── mod.rs
│   │   ├── open_button.rs
│   │   └── types
│   │       ├── mod.rs
│   │       └── sequence_load_state_map.rs
│   └── types
│       ├── camera_angle.rs
│       ├── character_pose_emotion.rs
│       ├── clip.rs
│       ├── image_filename_object.rs
│       ├── mod.rs
│       ├── page.rs
│       ├── pixel_size.rs
│       ├── router_context.rs
│       ├── sequence.rs
│       ├── time.rs
│       └── track.rs
└── lib.rs
```
</p>
</details>

luda-editor/luda-editor-client/src/app/app.rs
```diff
@@ -0,0 +1,122 @@
...
+            router: Router::new(RouterContext {
+                screen_size: namui::Wh {
+                    width: screen_size.width as f32,
+                    height: screen_size.height as f32,
+                },
+                socket,
+            }),
+        }
+    }
+    fn create_socket() -> luda_editor_rpc::Socket {
```
bigfoodK, line 33: `luda_editor_rpc::Socket` will be created at App

luda-editor/luda-editor-client/src/app/router.rs
```diff
@@ -0,0 +1,57 @@
...
+            Page::Editor(editor) => editor.update(event),
+            Page::SequenceList(sequence_list) => sequence_list.update(event),
+        }
+    }
+
+    fn render(&self, props: &Self::Props) -> namui::RenderingTree {
+        match &self.page {
+            Page::Editor(editor) => editor.render(props),
+            Page::SequenceList(sequence_list) => sequence_list.render(props),
+        }
```
bigfoodK, line 38: If all `page.render`s have the same prop, We can use page like `current_page: Option<namui::Entity<Props>>`. Then we don't need to use match.

luda-editor/luda-editor-client/src/app/events.rs
```diff
@@ -0,0 +1,10 @@
+use super::{editor::Editor, sequence_list::SequenceList, types::RouterContext};
+
+type PageInitializer<T> = Box<dyn Fn(&RouterContext) -> T + Send + Sync>;
+
+pub enum RouterEvent {
+    PageChangeToEditorEvent(PageInitializer<Editor>),
+    // TODO: this gonna be used to make back button in editor
+    #[allow(dead_code)]
+    PageChangeToSequenceListEvent(PageInitializer<SequenceList>),
+}
```
bigfoodK, line 10: Page instance will be created with this

luda-editor/luda-editor-client/src/app/app.rs
```diff
@@ -0,0 +1,122 @@
+use wasm_bindgen_futures::spawn_local;
+
```
namse, line 2: @bigfoodK There are some cases that put additional line break between `use`. Please remove that line breaks like this.

luda-editor/luda-editor-client/src/app/types/router_context.rs
```diff
@@ -0,0 +1,7 @@
+use luda_editor_rpc::Socket;
+use namui::Wh;
+
+pub struct RouterContext {
+    pub screen_size: Wh<f32>,
```
namse, line 5: I think it is not a context of router, but app's context. Would you rename it as `AppContext`?

luda-editor/luda-editor-client/src/app/types/camera_angle.rs
```diff
@@ -73,3 +75,9 @@ impl CameraAngle {
         )
     }
 }
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub struct Circumscribed {
```
namse, line 80: It's not a good position to put Circumscribed. Please leave only camera angle related thing in camera_angle.rs. Circumscribed should be in common like app/types/mod.rs or app/types/circumscribed.rs

luda-editor/luda-editor-client/src/app/types/clip.rs
```diff
@@ -0,0 +1,36 @@
...
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SubtitleClip {
+    pub id: String,
+    pub start_at: Time,
+    pub subtitle: Subtitle,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Subtitle {
```
namse, line 33: I have no idea that it is the good to put Subtitle struct in here. You moved structs to each specific files from mod.rs, some structs went to very specific file but some structs (like Subtitle or Circumscribed) just left in file which contains that struct. 
Is Subtitle struct related to clip? Is there dependency between clip and Subtitle? Clip needs Subtitles but Subtitles doesn't.  If I make a code which uses Subtitle, but not clip, I should import clip.rs because clip has subtitle even if I don't use clip. I think it's weird. 

I suggest making common file to put types about editor like editor_common.rs or make each specific file like subtitle.rs and circumscribed.rs. Personally I think the first one is better than seconds because this PR is about router, not a type files, and first idea keep the same type file system.

namse (changes requested):

luda-editor/luda-editor-client/src/app/types/clip.rs
bigfoodK: > I should import clip.rs because clip has subtitle even if I don't use clip.

Actually, thanks to re-export in mod.rs, we can use it with just via importing types like `App::types::Subtitle`.
But yes, `Subtitle` and `Circumscribed` are in wrong place now.

> Personally I think the first one is better than seconds because this PR is about router, not a type files,

I agree. I will rollback and do it later


c698ce0

luda-editor/luda-editor-client/src/app/types/router_context.rs
bigfoodK: Done with fa990a9

luda-editor/luda-editor-client/src/app/app.rs
bigfoodK: Done with ffb49c2

luda-editor/luda-editor-client/src/app/editor/clip_editor/camera_clip_editor/image_browser/mod.rs
```diff
@@ -1,14 +1,13 @@
...
-use wasm_bindgen_futures::spawn_local;
+use std::collections::BTreeSet;
 mod browser_item;
 use browser_item::*;
 mod back_button;
-use back_button::*;
 mod scroll;
 use scroll::*;
 
+use crate::app::{editor::events::EditorEvent, types::ImageFilenameObject};
```
namse, line 9: here remove line break please

luda-editor/luda-editor-client/src/app/editor/clip_editor/camera_clip_editor/wysiwyg_editor/mod.rs
```diff
@@ -1,8 +1,12 @@
+use crate::app::{
+    editor::{events::EditorEvent, job::Job},
+    types::*,
+};
+
```
namse, line 5: here remove line break please

luda-editor/luda-editor-client/src/app/editor/timeline/timeline_body/track_body/subtitle_track_body/mod.rs
```diff
@@ -1,12 +1,10 @@
...
-        SubtitleClipBody, SubtitleClipBodyProps,
-    },
-    types::{SubtitleClip, SubtitleTrack, Track},
-    Timeline, TimelineRenderContext,
+use crate::app::{
+    editor::{job::Job, TimelineRenderContext},
+    types::SubtitleTrack,
 };
 use namui::prelude::*;
+
```
namse, line 6: here remove line break please

namse (changes requested):

luda-editor/luda-editor-client/src/app/editor/mod.rs
```diff
@@ -204,8 +201,7 @@ impl namui::Entity for Editor {
 }
 
 impl Editor {
-    fn new(screen_wh: namui::Wh<f32>) -> Self {
-        let socket = Editor::create_socket();
+    pub fn new(screen_wh: namui::Wh<f32>, socket: Socket, sequence: Sequence) -> Self {
```
namse, line 204: I `think screen_wh` should not be a state of Editor, but just a props. I will make it clear what kind of data should be in entity struct and props later.

bigfoodK: Done with 4984bf0

PR #56, opened 2022-01-05
Optimize path by PathBuilder

namse:
# What happened?
It's very expensive job to make a canvaskit path. I found it's very nature in skia (https://stackoverflow.com/questions/15039829/drawing-paths-and-hardware-acceleration)
I faced very low fps situation when I create 10\~20 rectangle. what? 10\~20 rectangle makes luda-editor slow...!

# What I did
I replace Path::new() with PathBuilder::new() in luda-editor. PathBuilder is new struct which contains the command how you want to make a path.
PathBuilder only changes into path when namui draw it in AnimationFrame.
And also, PathBuilder has LRU Cache, it reuse path which it made before if the command is the same.

# How much it optimized namui?
Before I optimize it, the fps was about 5\~10 with 10\~20 rectangle in release, debug mode both.
After I optimize, in release mode, it was 58~60fps. It's pretty enough and good.
I got a low fps in debug mode but I think it's because of other element of canvaskit like paint or text, image, I drew it together. I will optimize it at next of this PR.

### DIFF
--- a/namui/src/namui/render/rect.rs
+++ b/namui/src/namui/render/rect.rs
@@ -47,37 +47,49 @@ pub fn rect(
     }: RectParam,
 ) -> RenderingTree {
     let mut rendering_tree: Vec<RenderingTree> = vec![];
-    let rect: namui::XywhRect<f32> = match stroke {
+    let (x, y, rect) = match stroke {
         None
         | Some(RectStroke {
             border_position: BorderPosition::Outside,
             ..
-        }) => XywhRect {
+        }) => (
             x,
             y,
-            width,
-            height,
-        },
+            XywhRect {
+                x: 0.0,
+                y: 0.0,
+                width,
+                height,
+            },
+        ),
         Some(RectStroke {
             border_position: BorderPosition::Middle,
             width: stroke_width,
             ..
-        }) => XywhRect {
-            x: x + stroke_width,
-            y: y + stroke_width,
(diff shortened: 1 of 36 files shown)


namui/src/namui/render/rect.rs
```diff
@@ -47,37 +47,49 @@ pub fn rect(
     }: RectParam,
 ) -> RenderingTree {
     let mut rendering_tree: Vec<RenderingTree> = vec![];
-    let rect: namui::XywhRect<f32> = match stroke {
+    let (x, y, rect) = match stroke {
```
namse, line 50: I extract x and y from rect and used `translate()` because of PathBuilder cache.

namui/src/namui/skia/base/path_builder.rs
```diff
@@ -0,0 +1,140 @@
...
+#[derive(Debug, Serialize, Clone)]
+pub struct PathBuilder {
+    commands: Vec<PathCommand>,
+}
+
+static PATH_CACHE: OnceCell<Mutex<lru::LruCache<PathBuilder, Arc<Path>>>> = OnceCell::new();
+
+impl PathBuilder {
+    pub(crate) fn build(&self) -> Arc<Path> {
+        self.get_or_create_path()
```
namse, line 33: You can get path by building PathBuilder.

PR #62, opened 2022-01-11
Set panic console log

namse:
I found that I didn't turn on panic console log. I REALLY DIDN'T KNOW THAT SO I THOUGHT WASM IS SUCKS... HAHAHAHAHAHAHAHAHAHAHAHAHAHA

### DIFF
--- a/luda-editor/luda-editor-client/Cargo.toml
+++ b/luda-editor/luda-editor-client/Cargo.toml
@@ -7,9 +7,6 @@ edition = "2021"
 [lib]
 crate-type = ["cdylib", "rlib"]
 
-[features]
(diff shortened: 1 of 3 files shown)


bigfoodK:
Just in case, is it intended to always turn on panic hook? Even if `target=release`

namse:
@bigfoodK I have no idea yet about that. Actually it's not a good case to turn on debug symbol things in release or production mode.
For now, it's little bit difficult to know target is release or not using [cfg](https://doc.rust-lang.org/rust-by-example/attribute/cfg.html). And I think we are not ready to prepare release mode including namui-cli. Let's see when we are preparing production mode.

PR #63, opened 2022-01-11
Change unreachable from todo

namse:
I think todo means what we will to do someday. I used todo! macro as unreachable but I think it is very different. So I replace some todo macros with unreachable.


bigfoodK:
I did't know there was `unreachable!`.
Good

PR #71, opened 2022-01-12
Enhance build server

### LINKED ISSUE #52: An unknown (0,0) appears when an error occurs during `namui -w.`
# What I did
1. make a rust coding error
2. run `namui -w`

# What I expected
1. Only coding error appears in chrome localhost:8080

# What happened
1. Unknown `(0,0)` appears together

image

### LINKED ISSUE #53: `namui-cli` report error, cannot refresh with save file.
I was coding, and a build error appeared in namui-cli. There are no errors in the file, but the error is still there.
It seems that there are cases where the rebuild does not work properly, need to check it.

image : Error message in namui-cli but no error in code.
image

bigfoodK:
# What I did
### 1. Exclude empty error message: 75abedc
This commit Fixes #52

#### What happened?
There was error while awaiting fs event.
`let (thread_sender, mut thread_receiver) = tokio::sync::mpsc::channel::<DebouncedEvent>(32);` 
After receiving a vast amount of events, `thread_receiver` doesn't receive anything. There was even no error.

#### How did I fixed it?
Instead of sending entire event between threads, send lighter data.
New data is `boolean` or 'tokio::sync::oneshot'

Tested with `git checkout master && sleep 0.1 && git checkout feature/enhance-build-server && sleep 0.1 && git checkout master && sleep 0.1 && git checkout feature/enhance-build-server && sleep 0.1 && git checkout master && sleep 0.1 && git checkout feature/enhance-build-server`

This commit Fixes #53 

### 1. Print debug message while cli_debug feature is on: 613415a
If run for build with `--features cli_debug` option, namui-cli will show you internal logs. Especially about lock(). And it will not clear console.
It will be SUPER ANNOYING.

### 2. Hide messages from wasm-pack: 22223b0
Many many warings in wasm-pack prevented me from debugging namui-cli. So I killed them 🥷.

### 1. Color file link blue for readability: d0bcddb
Error message and file path were the same color. This was confusing.
So color file path blue for readability.

### 2. Watch local dependency's path: aca7ebd
Watch not only the root src path but also the local dependency's src path.

It does not watch cargo.toml. But every after rebuild namui-cli will parse cargo.toml And then updates watching src list


namui-cli/src/build/build.rs
```diff
@@ -59,7 +54,7 @@ fn print_build_result(error_messages: &Vec<ErrorMessage>, cli_error_messages: &V
     println!("Errors {}", error_messages.len() + cli_error_messages.len());
     for error_message in error_messages {
         println!(
-            "{}\n\t--> {}:{}:{}\n",
+            "{}\n\t\x1b[34m--> {}:{}:{}\x1b[0m\n",
```
bigfoodK, line 57: `\x1b[34m` is command that setting color.
`\x1b[0m` is command that clearing color.

namui-cli/src/build/build_server/code_watcher.rs
```diff
@@ -0,0 +1,167 @@
...
+        match self.check_changed_from_last_check().await {
+            true => (),
+            false => {
+                let receiver = self.register_sender().await;
+                match receiver.await {
+                    Ok(_) => (),
+                    Err(error) => panic!("code_watcher oneshot receive error: {:?}", error),
+                }
+            }
+        }
```
bigfoodK, line 88: If there was changed after last check. It will return `()` immediately.
Else, It will wait for code change.

Returning `()` immediately means "There was code change while you rebuilding!"

namui-cli/src/build/build_server/code_watcher.rs
```diff
@@ -0,0 +1,167 @@
...
+use cargo_metadata::MetadataCommand;
+use futures::{executor::block_on, lock::Mutex};
+use notify::{DebouncedEvent, INotifyWatcher, RecursiveMode, Watcher};
+use regex::Regex;
+use std::{collections::HashSet, path::PathBuf, sync::Arc, thread, time::Duration};
+use tokio::sync::oneshot::{self, Sender};
+
+struct CodeWatcherContext {
+    was_changed: bool,
+    sender: Option<Sender<()>>,
```
bigfoodK, line 11: It looks queue but only one loop uses `CodeWatcher`.
So It will not be array of senders, but will be only one sender as optional

namse:
@bigfoodK What about hiding console outputs of cargo check here like wasm-pack?
I think it is unnecessary because namui-cli will clear console and show what's the error.
it's also a little bit laggy to print many lines in vscode terminal.

https://github.com/NamseEnt/namseent/pull/71/files#diff-7e2cda18f16243aed43d204902f579ffb42519fe79de1699e31508d4c615dbfcR29

namui-cli/src/build/build_server/code_watcher.rs
namse: I think it could make a timing issue. What if other thread found changes between 79 line and 82 line?

namui-cli/src/build/build_server/code_watcher.rs
namse: Is `only one loop` mean `only one consumer`?
If it's right, why don't you just make a callback parameter and just call it? because it's very simple and easy to debug.

namse (changes requested):

bigfoodK:
> @bigfoodK What about hiding console outputs of cargo check here like wasm-pack? I think it is unnecessary because namui-cli will clear console and show what's the error. it's also a little bit laggy to print many lines in vscode terminal.
> 
> https://github.com/NamseEnt/namseent/pull/71/files#diff-7e2cda18f16243aed43d204902f579ffb42519fe79de1699e31508d4c615dbfcR29

Good.
Added `println!("File changed. starting rebuild...");` to let user know that rebuilding has started,
5417578

namui-cli/src/build/build_server/code_watcher.rs
bigfoodK: Oh!. thx!
Timing issue is likely to be fixed by this commit 23ca59f.

namui-cli/src/build/build_server/code_watcher.rs
bigfoodK: If we use callback, we will probably call callback in watcher thread.
It will definitely be simpler than now.

But there's something I'm worried about.
If we block watcher thread, fs events will accumulate.
I think it's better to consume the event as soon as possible.
This is why I process events in another thread, the watcher thread.

But if you can rebuild on the main thread while using callback, that's probably a better way.
If you know how, can you recommend it to me?

namui-cli/src/build/build_server/code_watcher.rs
```diff
@@ -0,0 +1,164 @@
...
+                match context.sender {
+                    Some(_) => unreachable!(),
+                    None => context.sender = Some(sender),
+                }
+                CodeWatcherChangeState::NotChanged(receiver)
+            }
+        }
+    }
+
+    pub fn update_watching_paths(&mut self) {
```
namse, line 113: I think it is kind of initialization, so it would be better to make it private function and call in `new`.

namui-cli/src/build/build_server/code_watcher.rs
```diff
@@ -0,0 +1,164 @@
...
+                                debug_println!("CodeWatcher: code_watcher.context locked");
+
+                                match context.sender.take() {
+                                    Some(sender) => {
+                                        if let Err(_) = sender.send(()) {
+                                            panic!("send failed in watcher")
+                                        }
+                                    }
+                                    None => {
+                                        context.was_changed = true;
```
namse, line 63: If there are 2 DebouncedEvent, and if `context.sender` is not None, this thread will take `context.sender` and send it, and will make context.was_changed true, so it build twice, isn't it?

namse (changes requested):

namui-cli/src/build/build_server/code_watcher.rs
namse: Ah I got it. It should be called if Cargo.toml is changed. ok.

namui-cli/src/build/build_server/code_watcher.rs
namse: I think code `while watcher_receiver.try_recv().is_ok() {}` can flush events in mpsc.

namui-cli/src/build/build_server/code_watcher.rs
namse: I think it's the trade-off. But how many events would be accumulated? Is it really a lot to be worried?

bigfoodK: Ok good I will test it

bigfoodK: It works!

namui-cli/src/build/build_server/code_watcher.rs
```diff
@@ -29,30 +28,33 @@ impl CodeWatcher {
...
-                            return;
-                        }
-                        _ => (),
-                    };
+    pub fn wait_for_change(&mut self) {
+        let was_changed_while_flushing_events = self.flush_events();
+        match was_changed_while_flushing_events {
+            true => (),
+            false => loop {
+                match self.watcher_receiver.recv() {
```
bigfoodK, line 39: Before start waiting, It flush event queue and if there are necessary events, It return `()` immediately

namse: I have a question. Why flush before waiting?

It still have a problem as I commented here https://github.com/NamseEnt/namseent/pull/71#discussion_r783639092

namui-cli/src/build/build_server/start_build.rs
```diff
@@ -29,19 +29,17 @@ pub struct StartBuildOption {
 pub async fn start_build<'a>(option: StartBuildOption) {
     let mut watcher = CodeWatcher::new(option.manifest_path.clone());
 
+    println!("starting first rebuild...");
```
namse, line 32: I think it is just 'build', not rebuild because it is first time :)

namse: You can make a loop like do-while loop to make it simpler.

```
loop {
  build;
  wait();
}
```

namse (changes requested):

namui-cli/src/build/build_server/code_watcher.rs
namse: We made a conclusion not to test how events are accumulated and how much memory they used because we can assume it should be smaller than our thought.

namse:
I rebase and merge this pr because it has various topic's commits. Good job and well done @bigfoodK !

PR #73, opened 2022-01-14
Box namui::event internally

namse:
Previously, we have to call `namui::event::send(Box::new(Event))` to send event to namui. It's toooooo anoying to put `Box::new` everytime.

I found that we don't have to put `Box::new` everytime, just pass event as `impl Any + Send + Sync` and just box it inside of send function.

### DIFF
--- a/namui/src/namui/event/mod.rs
+++ b/namui/src/namui/event/mod.rs
@@ -24,3 +24,20 @@ pub enum NamuiEvent {
     ScreenResize(namui::Wh<i16>),
     Wheel(namui::Xy<f32>),
 }
+
+#[cfg(test)]
+mod tests {
+    #[tokio::test]
+    #[wasm_bindgen_test::wasm_bindgen_test]
+    async fn received_event_should_be_able_to_downcast() {
+        let mut event_receiver = super::init();
+        #[derive(Debug, PartialEq)]
+        enum Event {
+            Test,
+        }
+        super::send(Event::Test);
+        let event = event_receiver.recv().await.unwrap();
+        let downcasted = event.downcast_ref::<Event>().unwrap();
+        assert_eq!(downcasted, &Event::Test);
+    }
+}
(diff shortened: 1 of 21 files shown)


namui/src/namui/event/mod.rs
```diff
@@ -12,8 +12,8 @@ pub fn init() -> mpsc::UnboundedReceiver<Event> {
     receiver
 }
 
-pub fn send(event: Event) {
-    EVENT_SENDER.get().unwrap().send(event).unwrap();
+pub fn send(event: impl Any + Send + Sync) {
+    EVENT_SENDER.get().unwrap().send(Box::new(event)).unwrap();
```
namse, line 16: HERE what i changed. other code is just applying new syntax.

namse:
Rebase please. I add 45a4f46.

namui/src/namui/font/load_sans_typeface_of_all_languages.rs
```diff
@@ -91,22 +91,25 @@ mod tests {
...
+                (FontWeight::_100, "ko/NotoSansKR-Thin.otf".to_string()),
+                (FontWeight::_300, "ko/NotoSansKR-Light.otf".to_string()),
+                (FontWeight::_400, "ko/NotoSansKR-Regular.otf".to_string()),
+                (FontWeight::_500, "ko/NotoSansKR-Medium.otf".to_string()),
+                (FontWeight::_700, "ko/NotoSansKR-Bold.otf".to_string()),
+                (FontWeight::_900, "ko/NotoSansKR-Black.otf".to_string()),
+            ]),
+        )]);
+
+        assert_eq!(font_file_url_map, answer,);
```
bigfoodK, line 106:
```suggestion
        assert_eq!(font_file_url_map, answer);
```
If trailing comma was not intended, why don't we change it like this?

PR #79, opened 2022-01-17
Support set_timeout

namse:
I put set_timeout in namui.

# How it works
- namui store callback in binary-min-heap (or priority queue), sorting by time user requested.
- namui pop callback from heap on event loop.
  - this means, current set_timeout's timing is not accurate. I think it's not necessary to support accurate set-timeout for now.


namse:
I found that there is a problem with namui::now() in test because test assume 'now' is constant but it's not.

namse:
This should be merged after https://github.com/NamseEnt/namseent/pull/81

PR #83, opened 2022-01-18
Support text background rendering

namse:
I restore text background rendering from comments. other text-rendering-things will be restored next time.

### DIFF
--- a/namui/src/namui/render/text.rs
+++ b/namui/src/namui/render/text.rs
@@ -1,4 +1,7 @@
-use crate::namui::{self, *};
+use crate::{
+    draw::text::{get_bottom_of_baseline, get_left_in_align},
+    namui::{self, skia::GlyphIds, *},
+};
 
 #[derive(Clone, Copy, Debug)]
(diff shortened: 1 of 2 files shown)


namui/src/namui/render/text.rs
```diff
@@ -179,98 +185,88 @@ fn draw_text(param: TextParam, font: Arc<Font>) -> DrawCommand {
...
+    let glyph_bounds = font.get_glyph_bounds(glyph_ids, Option::None);
 
-//     let glyphs_top = glyph_bounds
-//         .iter()
-//         .reduce(|prev, cur| if prev.top < cur.top { prev } else { cur })
-//         .unwrap()
-//         .top;
+    let glyphs_top = glyph_bounds
+        .iter()
+        .reduce(|prev, cur| if prev.top < cur.top { prev } else { cur })
```
bigfoodK, line 261: I think it's reversed.

bigfoodK: Oh! It's distance from origin.

Issue #27, opened 2021-11-12
[TODO] Sequence List View


namse:
Yes, we have timeline. It's not about timeline.
I think timeline is the editor which can edit only single sequence.
We need some window or view which can select a sequence from sequence list.

# What result I am expecting
- New user experience progress for editor
  - Create new sequence
    - click 'add' button
    - type 'title'
    - good.
  - Editing sequence
    1. Showing sequence lists
    2. click one sequence
    3. open timeline editor
    4. edit, save, go back
    5. goto 1;
- Sequence List View
  - Title(required) and Thumbnail(optional)
  - Thumbnail is the first image of sequence? I have no idea.
- The way to fix title of sequence.

namse:
FYI, This commit implement the way to save and load sequence data. https://github.com/NamseEnt/namseent/commit/2ab54cbf072107f7fa064ecf7fd5b7c0448eb9a6

bigfoodK:
### TODO
- [x] List sequence titles
- [x] Open sequence
- [ ] Top bar(back to list button) in editor
- [ ] Create sequence
- [ ] Rename sequence
- [ ] Delete sequence

bigfoodK:
Closed with #88, #95

PR #88, opened 2022-01-20
List sequences and let them open

bigfoodK:
### Features
- List sequence titles
- Open sequence

It is part of #27 
I opened this pr for easier code reviews.

image


luda-editor/luda-editor-client/src/app/sequence_list/list_item.rs
```diff
@@ -0,0 +1,84 @@
...
+                                        path,
+                                        state: None,
+                                    },
+                                );
+                            }
+                            false => {
+                                let path = path.clone();
+                                namui::event::send(SequenceListEvent::SequenceLoadEvent { path });
+                            }
+                        }))
```
bigfoodK, line 42: Is there a good way to remove `let path = path.clone();`?

luda-editor/luda-editor-client/src/app/sequence_list/mod.rs
```diff
@@ -108,31 +121,103 @@ impl Entity for SequenceList {
...
+                                                        error,
+                                                    },
+                                            },
+                                        },
+                                    );
+                                }
+                            }
+                        }
+                    });
+                }
```
bigfoodK, line 194: Same as `SequenceLoadEvent` and `SequenceLoadStateUpdateEvent`

luda-editor/luda-editor-client/src/app/sequence_list/list.rs
```diff
@@ -0,0 +1,96 @@
...
+                .attach_event(move |builder| {
+                    let scroll_y = clamped_scroll_y;
+                    builder.on_wheel(Box::new(move |event| {
+                        let delta_y = event.delta_xy.y;
+                        let next_scroll_y = clamp(scroll_y + delta_y, 0.0, max_scroll_y);
+                        namui::event::send(SequenceListEvent::ScrolledEvent {
+                            scroll_y: next_scroll_y,
+                        });
+                    }))
+                }),
```
bigfoodK, line 68: Wherever cursor is, the scroll will works.

luda-editor/luda-editor-client/src/app/sequence_list/events.rs
```diff
@@ -8,4 +8,11 @@ pub enum SequenceListEvent {
     SequenceLoadEvent {
```
namse, line 8: It would be better to name it like `SequenceLoadRequestedEvent` because when I read this event, I thought it's the event about loading, but it seems to make start loading sequence.

Personally I tried to make a event name just what exactly happened, not as what I have to do.
For example, I would name it like `SequenceTitleButtonClickEvent`, because user click the button. It's not their business to load sequence or not. I just let upper-side know that I click this button, then they will handle it, and they will give me next props, it just render about props.

luda-editor/luda-editor-client/src/app/sequence_list/mod.rs
```diff
@@ -1,35 +1,48 @@
...
+const MARGIN: f32 = 4.0;
 
 pub struct SequenceListProps {
-    pub xywh: XywhRect<f32>,
+    pub wh: Wh<f32>,
 }
 
 pub struct SequenceList {
     sequence_load_state_map: SequenceLoadStateMap,
+    sequence_titles_load_state: Option<SequenceTitlesLoadState>,
```
namse, line 33: I think it would not to be `Option` because `SequenceTitlesLoadState` has loading state, and you start load on `SequenceList::new`.

luda-editor/luda-editor-client/src/app/sequence_list/types/sequence_titles_load_state.rs
```diff
@@ -0,0 +1,14 @@
...
+#[derive(Clone)]
+pub enum SequenceTitlesLoadStateDetail {
+    Loading,
+    Loaded { titles: Vec<String> },
+    Failed { error: String },
+}
+
+#[derive(Clone)]
+pub struct SequenceTitlesLoadState {
+    pub started_at: Duration,
```
namse, line 12: I think it's kind of over-engineering to handle `started_at` for loading sequence titles, because it only called once, not multiple time at once. How do you think about it?

luda-editor/luda-editor-client/src/app/sequence_list/rounded_rectangle.rs
```diff
@@ -0,0 +1,50 @@
...
+
+impl convert::Into<Color> for RoundedRectangleColor {
+    fn into(self) -> Color {
+        match self {
+            RoundedRectangleColor::DarkGray => Color::grayscale_f01(0.3),
+            RoundedRectangleColor::Gray => Color::grayscale_f01(0.5),
+            RoundedRectangleColor::LightGray => Color::grayscale_f01(0.8),
+            RoundedRectangleColor::Blue => Color::from_u8(107, 185, 240, 255),
+            // RoundedRectangleColor::Red => Color::from_u8(242, 38, 19, 255),
+            // RoundedRectangleColor::White => Color::WHITE,
```
namse, line 47: I think you can remove this comment or uncomment.

luda-editor/luda-editor-client/src/app/sequence_list/mod.rs
```diff
@@ -108,31 +121,103 @@ impl Entity for SequenceList {
...
+                                            },
+                                        },
+                                    );
+                                }
+                            }
+                        }
+                    });
+                }
+                SequenceListEvent::ScrolledEvent { scroll_y } => {
+                    self.scroll_y = *scroll_y;
```
namse, line 196: You should make sure that mouse is on sequence list or not. It's ok for now because there are nothing but just sequence list in screen. You don't have to fix it now but keep in mind about it.
Reference : https://github.com/NamseEnt/namseent/blob/master/luda-editor/luda-editor-client/src/app/editor/clip_editor/camera_clip_editor/image_browser/scroll.rs#L129

luda-editor/luda-editor-client/src/app/sequence_list/mod.rs
```diff
@@ -108,31 +121,103 @@ impl Entity for SequenceList {
...
+                SequenceListEvent::SequenceTitlesLoadStateUpdateEvent { state } => {
+                    if let Some(old_state) = &self.sequence_titles_load_state {
+                        if old_state.started_at > state.started_at {
+                            return;
+                        }
+                    }
+
+                    self.sequence_titles_load_state = Some(state.clone());
+                }
+                SequenceListEvent::SequenceTitlesLoadEvent => {
```
namse, line 133: I think it's too big code to put in `update`. Would you like to move this logic into other fn with nice name and just call it?

luda-editor/luda-editor-client/src/app/sequence_list/list_item.rs
```diff
@@ -0,0 +1,84 @@
...
+            render![
+                self.render_rounded_rectangle(button_wh, RoundedRectangleColor::DarkGray)
+                    .with_mouse_cursor(namui::MouseCursor::Pointer)
+                    .attach_event(move |builder| {
+                        let path = path.clone();
+                        builder.on_mouse_down(move |_| match is_open {
+                            true => {
+                                let path = path.clone();
+                                namui::event::send(
+                                    SequenceListEvent::SequenceLoadStateUpdateEvent {
```
namse, line 32: Personally, I found it very difficult to read the code here. it maybe because of the name of event or variables.

`if is open` <- has no subject. So when I read code, I thought about 'what can be opened'. 'opening file'? 'opening... what?'

I really found what I can open when I run the demo. AH. The list inside of button! HOW CAN I KNOW THAT only reading this code :S

the code must be nicer if it like this,
```
if list(or inner_list) is open {
  close it()
} else {
  open it()
}
fn close {
  // put what I have to do to close
}
fn open {
  // put what I have to do to open
  start load()
}
```

It's very important to divide code into functions because when we read the code, it's not our business what exactly your code do.
It's important how the logic is flowing. That mean the code should hide the implementation, just show the interaction between interfaces, functions with good name to feel the flow.
After read all flow, I read the implementation. It's lower priority. logic is more important than implementation, that mean we have to write the code to let programmer know how the logic is.

So anyway, would you like to fix this code more readable?

luda-editor/luda-editor-client/src/app/sequence_list/list_item.rs
```diff
@@ -0,0 +1,84 @@
...
+
+impl SequenceList {
+    pub fn render_list_item(&self, width: f32, title: &String) -> RenderingTreeRow {
+        let path = format!("sequence/{}", title);
+        let element_width = width - 2.0 * MARGIN;
+        let button_wh = Wh {
+            width: element_width,
+            height: BUTTON_HEIGHT,
+        };
+        let title_load_state = self.sequence_load_state_map.get(&path);
```
namse, line 18: I have no idea that it's the good idea to know whether the list is open or not using `sequence_load_state_map`.
`sequence_load_state_map` is the map to check about sequence loading state.
But you used it to determine whether the list is open or not in here.
One data for two purpose would make code complex.

I believe you at least hide this in function and manage that function. or just get information from parameter of `render_list_item`.

Do you have any idea to improve this?

namse (changes requested):

luda-editor/luda-editor-client/src/app/sequence_list/list_item.rs
namse: You cannot remove 'clone' for move String to closure because String didn't implement Copy trait.  [Reference](http://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/std/marker/trait.Copy.html#when-cant-my-type-be-copy)

you used `let path = path.clone()` 3 times, but it could be 2 times.

```rust
                        let path = path.clone();
                        builder.on_mouse_down(Box::new(move |_| {
                            let path = path.clone();
                            match is_open {
                                true => {
                                    namui::event::send(
                                        SequenceListEvent::SequenceLoadStateUpdateEvent {
                                            path,
                                            state: None,
                                        },
                                    );
                                }
                                false => {
                                    namui::event::send(SequenceListEvent::SequenceLoadEvent { path });
                                }
                            }))
```

luda-editor/luda-editor-client/src/app/sequence_list/list.rs
namse: Is it intended?

luda-editor/luda-editor-client/src/app/sequence_list/list_item.rs
namse: It's good if the code inside doesn't know the outside. Problems arise the moment you know the outside from the inside.

luda-editor/luda-editor-client/src/app/sequence_list/list.rs
bigfoodK: Yes.
Because there's no other entity that uses scrolling right now

luda-editor/luda-editor-client/src/app/sequence_list/events.rs
bigfoodK: I think both of them would be better than now.
But I don't know which one is better between `SequenceLoadRequestedEvent` and `SequenceTitleButtonClickedEvent`.

I'll choose `SequenceTitleButtonClickedEvent` for the consistency of the code.

luda-editor/luda-editor-client/src/app/sequence_list/mod.rs
bigfoodK: Yes, that must be lagacy code. I'll fix them

luda-editor/luda-editor-client/src/app/sequence_list/types/sequence_titles_load_state.rs
bigfoodK: I think user can click `Reload titles` button multiple time at once.

luda-editor/luda-editor-client/src/app/sequence_list/rounded_rectangle.rs
bigfoodK: It will be used to make features in issue.
But if I don't comment it. It shows me `dead_code` warning. It was very annoying.
So I commented it and I'll uncomment it when I use it

luda-editor/luda-editor-client/src/app/sequence_list/mod.rs
bigfoodK: How about `fn handle_*_event()`?
Does it sound like a bad name?

luda-editor/luda-editor-client/src/app/sequence_list/mod.rs
bigfoodK: Fixed with 3d15814

luda-editor/luda-editor-client/src/app/sequence_list/events.rs
bigfoodK: Fixed with 7166c1e

luda-editor/luda-editor-client/src/app/sequence_list/mod.rs
bigfoodK: `fn handle_*_event()` was bad name I think.
Changed to name that represents what a function does with 4653fbf

bigfoodK:
### Changes
- Change event naming: 7166c1e
- Split large update function into another function: 4653fbf
- Render elements with function instead of impls: 15bc772
  - prev: `self.render_something()`
  - now: `render_something()`
- Make sequence_titles_load_state non-optional: 3d15814
  - SequenceList will load titles on init. sequence_titles_load_state always be non-optional

namse:
Wow... Your code finally has really really good readability! Even though I read this pr code twice, but this time it was really easy to understand how it works. good job!

namse (approved):

PR #89, opened 2022-01-20
Get buttons on mouse event

namse:
# What I did
I put buttons in mouse event.

# How buttons work
Exactly the same with web's MouseEvent.buttons. https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons

### DIFF
--- a/namui/src/namui/common/mod.rs
+++ b/namui/src/namui/common/mod.rs
@@ -4,6 +4,7 @@ use super::skia::*;
 use crate::event::EventReceiver;
 use serde::{Deserialize, Serialize};
 use serde_repr::*;
+use std::collections::HashSet;
 use std::time::Duration;
 use strum_macros::EnumIter;
 mod xy;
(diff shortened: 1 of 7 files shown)


namui/src/namui/common/mod.rs
```diff
@@ -173,3 +174,15 @@ pub struct TypefaceType {
...
+#[derive(Debug, Hash, Eq, PartialEq, Clone, Copy)]
+pub enum MouseButton {
+    Left,
+    Middle,
+    Right,
+}
+
+pub struct RawMouseEvent {
+    pub xy: Xy<f32>,
+    pub buttons: HashSet<MouseButton>,
```
namse, line 187: If HashSet is too heavy, I can make light struct for it using bit-wise operator. But I don't think that we face the performance issue right now.

bigfoodK (commented):
LGTM

bigfoodK (approved):
LGTM

PR #92, opened 2022-01-25
Move timeline event handling from editor

namse:
It's simple pull request.

I found that Editor handle some event about only for timeline. it even change timeline like `self.timeline.xxx = yyy`.
Hmm. I think it should be in timeline code. so I moved it from editor to timeline.


bigfoodK (approved):
Good

PR #106, opened 2022-02-03
Recover from fetch fail error

namse:
# What happened
I thought our ImageManager can recover from error of fetching image. But it doesn't, it become panic.
I found that fetch logic used unwrap, so it was not recovered even it's return type was Result.

# What I did
I changed unwrap-ing code into question mark operator, which return Err instead of panic.

### DIFF
--- a/namui/src/namui/namui_web/fetch.rs
+++ b/namui/src/namui/namui_web/fetch.rs
@@ -1,52 +1,69 @@
+use core::fmt;
+
 use js_sys::{ArrayBuffer, Uint8Array};
-use wasm_bindgen::JsCast;
+use wasm_bindgen::{JsCast, JsValue};
 use wasm_bindgen_futures::JsFuture;
 use web_sys::{Request, RequestInit, Response};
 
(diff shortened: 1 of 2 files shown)


namui/src/namui/namui_web/fetch.rs
```diff
@@ -1,50 +1,69 @@
...
-    let array_buffer = fetch_get_array_buffer(url).await.unwrap();
+pub async fn fetch_get_vec_u8(url: &str) -> Result<Vec<u8>, FetchError> {
+    let array_buffer = fetch_get_array_buffer(url).await?;
     let array_buffer_view = Uint8Array::new(&array_buffer);
     let bytes = array_buffer_view.to_vec();
     Result::Ok(bytes)
 }
+
+#[derive(Debug)]
+pub struct FetchError {
```
namse, line 53: Most of Err type in this code is JsValue. so it's not possible to implement `From<String> for String` because that struct is not defined by our crate. 
So I created new struct and wrap errors with it.

bigfoodK (approved):

PR #108, opened 2022-02-04
Show camera clip preview on timeline

namse:
# What was the problem
In the timeline, it was very difficult to distinguish the camera clips. Because all the clips were the same color and the same shape.

# How did I solve it
I show camera clip preview in camera clip body, like other video editing tools.

# Design Doc
https://docs.google.com/document/d/1bs7z9n0Oy0qpS-00k946Yz69BYClqtIpL5pBrh10HLA/edit#

### DIFF
--- a/luda-editor/luda-editor-client/src/app/types/camera_angle.rs
+++ b/luda-editor/luda-editor-client/src/app/types/camera_angle.rs
@@ -73,35 +73,17 @@ impl CameraAngle {
             height: image_size.height,
         };
 
-        let background = rect(RectParam {
-            x: 0.0,
-            y: 0.0,
-            width: wh.width,
-            height: wh.height,
-            style: RectStyle {
-                fill: Some(RectFill {
-                    color: Color::WHITE,
-                }),
-
-                ..Default::default()
-            },
-            ..Default::default()
-        });
-
-        render![
-            background,
-            clip(
-                PathBuilder::new().add_rect(&clip_rect),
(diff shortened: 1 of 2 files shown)


luda-editor/luda-editor-client/src/app/types/camera_angle.rs
```diff
@@ -73,17 +73,35 @@ impl CameraAngle {
...
-            namui::image(ImageParam {
-                source: ImageSource::Image(image),
-                xywh: image_xywh,
-                style: ImageStyle {
-                    fit: ImageFit::Fill,
-                    paint_builder: None,
-                },
-            }),
-        )
+        let background = rect(RectParam {
```
namse, line 76: I found that CameraAngle::render doesn't have background. So I put simple white background here.

bigfoodK: I think it's better to draw background in [camera_clip_body.rs](https://github.com/NamseEnt/namseent/pull/108/files#diff-c4a608b20e52512e869c2fe29be048a67a3d1efc6ca8dbbe36bdbc2e8bca889b) than in [camera_angle.rs](https://github.com/NamseEnt/namseent/pull/108/files#diff-635a3307a31f5ef5ba090296fa5df5cb72891169da166efa83c282e5e33a57fd).
Because the background is only needed for the clip body.

bigfoodK (approved):

PR #115, opened 2022-02-08
Show flag on subtitle clip to check to update

namse:
I rendered 'Check!' text on subtitle clip if
1. it is new subtitle clip
2. it is not very new, but position moved by syncing algorithm


bigfoodK:
It looks good and works well.
But I have a question. If user runs sync twice, what should happen to unchecked clips? 
Should they be switched to checked clips even if user didn't modify them?
Currently, when sync is executed twice, all clips are switched to checked.

namse:
> It looks good and works well. But I have a question. If user runs sync twice, what should happen to unchecked clips? Should they be switched to checked clips even if user didn't modify them? Currently, when sync is executed twice, all clips are switched to checked.

You are very right, I added about that case as test case and fixed it --> https://github.com/NamseEnt/namseent/pull/115/commits/119b4e52e042156398bb7308cc5a6df169426d2c

bigfoodK (approved):

Issue #82, opened 2022-01-18
`namui-cli` is stuck at rebuilding


namse:
![image](https://user-images.githubusercontent.com/3580430/149932604-8c77eac0-fcec-4f5e-a392-0f020a8bd58a.png)

(1) - Every file are saved that I changed.
(2) - But namui-cli is stuck at rebuilding

![image](https://user-images.githubusercontent.com/3580430/149932801-2262a4d0-144a-458e-b73f-2e2a13cfa02f.png)

`cargo check` is working.

It become fine when I re-run namui-cli.

namse:
Not happened nowadays.

Issue #70, opened 2022-01-12
Namui-cli doesn't send bundle.


bigfoodK:
Today, I face to simillar error that namui doesn't rebuild on changes and doesn't send bundle neither.

https://user-images.githubusercontent.com/3580430/148041885-1aedf9c4-0d32-4de5-a501-9027729862a7.mp4

_Originally posted by @namse in https://github.com/NamseEnt/namseent/issues/53#issuecomment-1004669803_

It looks another problem from #53

namse:
It isn't reproduced for now. Let's reopen when it happen again.

PR #132, opened 2022-02-26
Use and_then to serve bundle

bigfoodK:
# What happened
An unknown stuck occurs in the `Bundle`.
Related issues: #70, #82
I couldn't find the definite conditions for stuck to occur. So I can't guarantee that this pr solves the above issues.

# What I did?
I replace `map` and `block_on` with `warp::Filter.and_then`.
This is not a common way to use async with warp. I didn't know how to use `warp::Filter.and_then` with local variables at that time. So I used trick of `map` and `block_on`.

In my case, since this modification, stuck has no longer occurred. So I made this pr for more tests.

### DIFF
--- a/namui-cli/src/build/web_server/web_server.rs
+++ b/namui-cli/src/build/web_server/web_server.rs
@@ -7,7 +7,6 @@ use crate::{
 };
 use futures::{
     channel::mpsc::{unbounded, UnboundedSender},
-    executor::block_on,
     lock::Mutex,
     SinkExt, StreamExt,
 };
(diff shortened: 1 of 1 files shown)


namui-cli/src/build/web_server/web_server.rs
```diff
@@ -38,30 +38,40 @@ impl WebServer {
...
 
         let bundle = option.bundle.clone();
-        let serve_wasm_bundle = warp::path("bundle_bg.wasm").map(
-            move || -> Result<warp::hyper::Response<Vec<u8>>, warp::http::Error> {
+        let serve_wasm_bundle = warp::path("bundle_bg.wasm").and_then(move || {
+            let bundle = bundle.clone();
+            async move {
                 debug_println!("serve_wasm_bundle: locking web_server.bundle...");
-                let bundle = block_on(bundle.read()).wasm.clone();
+                let wasm = block_on(bundle.read()).wasm.clone();
```
namse, line 45: I think this block_on should be changed like below, the js part.

namse (changes requested):

namse (approved):

PR #136, opened 2022-03-07
Refactor namui-cli

namse:
In shorts, this pr is the pre-pr about supporting wasm-electron. before adding wasm-electron, I refactored namui-cli to be more flexible.

# What I changed
- Removed `-w`
  - I will put other flag something like `namui wasm-web --dev` or `namui wasm-electron --release`, but more familiar with cargo commands.
- Services and Procedure 
  - Procedure is kind of recipe.
  - Services is common part that can be reused among Procedures.
- Renewal or simplified about services
- Removed www folder in namui and projects.

Related design docs : https://docs.google.com/document/d/1FeSWV6WAeeOjuAmahLNorA5Adj6X7dpIVg3EqVgpcVI/edit#


namui-cli/src/services/rust_build_service.rs
```diff
@@ -0,0 +1,266 @@
...
+
+impl RustBuildService {
+    pub(crate) fn new() -> Self {
+        Self {
+            builder: Mutex::new(None),
+            is_build_just_started: AtomicBool::new(false),
+        }
+    }
+
+    pub(crate) fn cancel_and_start_build(&self, build_option: &BuildOption) -> BuildResult {
```
namse, line 45: From now on, namui-cli's build system support cancellation.

namui-cli/src/services/rust_build_service.rs
```diff
@@ -0,0 +1,266 @@
...
+            return BuildResult::Canceled;
+        }
+
+        let result_receiver = {
+            let mut builder_lock = self.builder.lock().unwrap();
+            if let Some(builder) = builder_lock.take() {
+                builder.cancel();
+            }
+
+            self.is_build_just_started.store(false, Ordering::Relaxed);
```
namse, line 56: Only one access can cancel the build, and during that time, other accesses will be canceled before trying to cancel the build.

namui-cli/src/services/rust_build_service.rs
```diff
@@ -0,0 +1,266 @@
...
+    pub fn cancel(&self) {
+        if self.is_canceled.load(Ordering::Relaxed) {
+            return;
+        }
+
+        self.is_cancel_requested.store(true, Ordering::Relaxed);
+        debug_println!("build cancel requested");
+
+        loop {
+            thread::sleep(Duration::from_millis(100));
```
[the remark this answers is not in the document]
namse: I failed to make it using mpsc channel.

namui-cli/src/services/rust_build_service.rs
```diff
@@ -0,0 +1,266 @@
...
+
+        let builder_thread_fn = {
+            move |builder: Arc<Self>| -> Result<BuildResult, Box<dyn std::error::Error>> {
+                let mut spawned_process = Self::spawn_build_process(&build_option)?;
+
+                let mut stdout = spawned_process.stdout.take().unwrap();
+                let mut stderr = spawned_process.stderr.take().unwrap();
+                let stdout_reading_thread = thread::spawn(move || {
+                    let mut string = String::new();
+                    stdout.read_to_string(&mut string).unwrap();
```
namse, line 110: there are many report about stuck the child process when they didn't flush outputs. it is blocking function so I ran thread for it.

ref: https://stackoverflow.com/questions/34611742/how-do-i-read-the-output-of-a-child-process-without-blocking-in-rust

namui-cli/src/services/rust_build_service.rs
```diff
@@ -0,0 +1,266 @@
...
+                }
+            }
+        };
+
+        let (result_sender, result_receiver) = std::sync::mpsc::channel();
+
+        thread::spawn({
+            let builder = builder.clone();
+            move || {
+                let build_result = match builder_thread_fn(builder.clone()) {
```
namse, line 159: this is for using `?` in builder_thread_fn.

namui-cli/src/services/wasm_bundle_web_server.rs
```diff
@@ -129,47 +91,58 @@ impl WebServer {
...
 
-        let _ = spawn(warp::serve(routes).run(([0, 0, 0, 0], option.port)));
+        let _ = tokio::spawn(warp::serve(routes).run(([0, 0, 0, 0], port)));
 
         web_server
     }
 
-    pub async fn send_error_messages(&self, error_messages: &Vec<ErrorMessage>) {
-        debug_println!("send_error_messages: locking web_server.sockets...");
+    pub async fn on_build_done(&self, result: &CargoBuildResult) {
```
namse, line 107: i just merged two function into this one.

namui-cli/src/util/debug_println.rs
```diff
@@ -8,5 +8,5 @@ macro_rules! debug_println {
 #[cfg(not(feature = "cli_debug"))]
 #[macro_export]
 macro_rules! debug_println {
-    ($( $args:expr ),*) => {};
+    ($( $args:expr ),*) => {$( let _ = $args; )*};
```
namse, line 11: This is very important. It make compiler check whether the args are good or not in not-cli_debug mode too.
Previously, this macro just ignored args so compiler didn't check about args.

bigfoodK (approved):

bigfoodK:
If we don't need clang in the next version that supports electron, clang should be deleted.
Remember!
https://github.com/NamseEnt/namseent/blob/2e52f8c9b7b5c1b10e7ff44db6405e1ae54e4693/namui-cli/Cargo.toml#L12

namse:
REMEMBER

PR #142, opened 2022-03-15
Calculate subtitle duration with specific text token

### LINKED ISSUE #140: Token based Subtitle Duration Calculation
# What Happened
When the subtitle contains `...` or `,`, it should make subtitle duration longer. 
It should has different duration between `...` and `,`.

So I got a needs to calculate subtitle duration in different way when it comes to with specific text token.

bigfoodK:
# What Happened
https://github.com/NamseEnt/namseent/issues/140#issue-1167625617
> When the subtitle contains `...` or `,`, it should make subtitle duration longer. It should has different duration between `...` and `,`.
> 
> So I got a needs to calculate subtitle duration in different way when it comes to with specific text token.

# What I Did
Add `subtitle_specific_text_token_play_duration_map` to `meta.json`.
You can add token like this
```json
{
    "subtitle_language_minimum_play_duration_map": {..},
    "subtitle_language_play_duration_per_character_map": {..},
    "subtitle_specific_text_token_play_duration_map": {
      "...": {
        "milliseconds": 200.0
      }
    },
    "subtitle_character_color_map": {..}
}
```

Now `subtitle_play_duration_measure` calculates duration like following
1. Iter tokens in `meta.subtitle_specific_text_token_play_duration_map`
  a. Find token in subtitle
  b. Add (token count * token duration) to subtitle duration
  c. Remove token from subtitle
2. Add (remaining subtitle * duration per character) to subtitle duration

Close #140

### DIFF
--- a/luda-editor/luda-editor-client/src/app/types/meta.rs
+++ b/luda-editor/luda-editor-client/src/app/types/meta.rs
@@ -2,6 +2,7 @@ use std::sync::Mutex;
 
 use super::*;
 use async_trait::async_trait;
+use linked_hash_map::LinkedHashMap;
 use luda_editor_rpc::Socket;
 use serde::{Deserialize, Serialize};
 use wasm_bindgen_futures::spawn_local;
@@ -10,6 +11,7 @@ use wasm_bindgen_futures::spawn_local;
 pub struct Meta {
     pub subtitle_language_minimum_play_duration_map: HashMap<Language, Time>,
     pub subtitle_language_play_duration_per_character_map: HashMap<Language, Time>,
+    pub subtitle_specific_text_token_play_duration_map: LinkedHashMap<String, Time>,
     pub subtitle_character_color_map: HashMap<String, Color>,
 }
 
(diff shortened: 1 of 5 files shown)


luda-editor/luda-editor-client/src/app/types/meta.rs
```diff
@@ -10,6 +11,7 @@ use wasm_bindgen_futures::spawn_local;
 pub struct Meta {
     pub subtitle_language_minimum_play_duration_map: HashMap<Language, Time>,
     pub subtitle_language_play_duration_per_character_map: HashMap<Language, Time>,
+    pub subtitle_specific_text_token_play_duration_map: LinkedHashMap<String, Time>,
```
bigfoodK, line 14: I used `LinkedHashMap` instead `HashMap` for ordering
`LinkedHashMap` keeps order of entry

luda-editor/resources/meta.json
```diff
@@ -9,6 +9,7 @@
             "milliseconds": 200.0
         }
     },
+    "subtitle_specific_text_token_play_duration_map": {},
```
bigfoodK, line 12: I just added empty object.
Please add as much as you want in other commit😊.

namse (approved):

PR #145, opened 2022-03-18
Install cli on windows

### LINKED ISSUE #144: Cli can not get cli root path on windows
# What happened?
I'm trying to use namui-cli on windows.
So I
1. Build namui-cli with `cargo build`
2. Make symlink with `New-item -ItemType SymbolicLink -Path $SymlinkPath -Target "$CliPath -test" -Force` in powershell
3. Run cli with `namui` in `namseent\luda-editor\luda-editor-client`

But cli serves wrong static path.
https://github.com/NamseEnt/namseent/blob/13d795857e872fe871fcbef5a105d1cb0bee8107/namui-cli/src/services/wasm_bundle_web_server.rs#L175
In here.  `current_exe` returns symlink's path, not namui-exe's path.

FYI
https://doc.rust-lang.org/std/env/fn.current_exe.html#platform-specific-behavior

bigfoodK:
# What I did
## Add install and uninstall script
Add `install.ps1` and `uninstall.ps1` to `namui-cli/scripts`
Install script should be executed with administrative privileges. 
Because of next line
https://github.com/NamseEnt/namseent/pull/145/files#diff-d2d04ddfc0aeb03d7a265cb12c138b5967ac4197cf5cbf48ea4d747f0c701b83R91

Install script does
- Check cargo installed
- Check wasm-pack installed
- Build cli
- Make symlink of namui-cli.exe to cargo bin dir
- Make config file next to symlink

Uninstall script does
- Remove symlink
- Remove config file

## Resolve #144
via making config file
`namui.config.json` contains path of real exe file, not symlink's

Close #144

### DIFF
--- a/namui-cli/scripts/install.ps1
+++ b/namui-cli/scripts/install.ps1
@@ -1,0 +1,114 @@
+function Main {
+    $CliRootPath = Get-CliRootPath
+    $CliPath = Get-CliPath -CliRootPath $CliRootPath
+    $CargoBinDirPath = Get-CargoBinDirPath
+
+    Assert-CargoInstalled
+    Assert-WasmPackInstalled
+    Assert-CargoBinDirExist -CargoBinDirPath $CargoBinDirPath
+
+    Build-Cli -CliRootPath $CliRootPath
+
+    New-CliSymlink -CliPath $CliPath -CargoBinDirPath $CargoBinDirPath
+
+    Initialize-Config -CliPath $CliPath -CargoBinDirPath $CargoBinDirPath
+
+    Write-Output "Successfully installed."
+}
+
+# Error Code
+$EXIT_CARGO_NOT_FOUND = 1
+$EXIT_WASM_PACK_NOT_FOUND = 2
+$EXIT_CARGO_BIN_DIR_NOT_FOUND = 3
+$EXIT_CLI_BUILD_FAILED = 4
+$EXIT_SYMLINK_MAKE_FAILED = 5
+$EXIT_CONFIG_INIT_FAILED = 6
+
+function Get-CliRootPath {
+    return (Get-Item $PSScriptRoot).Parent.FullName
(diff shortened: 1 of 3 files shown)


namui-cli/scripts/install.ps1
```diff
@@ -0,0 +1,114 @@
...
+function Initialize-Config {
+    param (
+        [string] $CliPath,
+        [string] $CargoBinDirPath
+    )
+
+    $ConfigJson = @{ exe_path = $CliPath } | ConvertTo-Json
+    $ConfigPath = Join-Path -Path $CargoBinDirPath -ChildPath "\namui.config.json"
+    $Utf8NoBomEncoding = New-Object System.Text.UTF8Encoding $False
+    [System.IO.File]::WriteAllLines($ConfigPath, $ConfigJson, $Utf8NoBomEncoding)
```
bigfoodK, line 107: `Out-File -Encoding "utf8"` makes file with bom, It causes error while reading json in rust.
So I did like this

bigfoodK: FYI https://stackoverflow.com/questions/5596982/using-powershell-to-write-a-file-in-utf-8-without-the-bom

namui-cli/src/services/wasm_bundle_web_server.rs
```diff
@@ -172,13 +175,32 @@ impl WasmBundleWebServer {
 }
 
 fn get_cli_root_path() -> PathBuf {
-    let mut exe_path = current_exe().expect("Current exe path not found.");
+    let mut exe_path = get_exe_path_from_config()
+        .or(current_exe())
```
bigfoodK, line 179: Try read config file first, then use `current_exe()` directly.

bigfoodK:
We don't install namui on Windows for a while.
So close this pr.

PR #149, opened 2022-03-31
Use human-friendly format for namui bundle index file

bigfoodK:
# What I did
Previously, we saved it in resources property of namui.config.json.
- But it's annoying to modify json directly.
- And we just need resources in config for now

So
- Use custom format that is human-friendly. It looks like .gitignore
- Remove `namui.config.json` and use `.namuibundle`.

# `.namuibundle` format
```
# comment, this sould be ignored

path/to/src
path/to/src:path/to/dest
```
- One item should be just one line
- Line starts with # will be ignored
- Empty line will be ignored


namse:
I think the word `index` confuse me because I usually use `index` for for-loop... i guess?
What about use `list` or plural? Index doesn't feel that there are many data, but just feel single data.

bigfoodK:
``# New features
## **: Test all directory below
`dir/**/file` on
```
dir/
├── dir1/
│   ├── file
│   └── dir3/
│       └── file
└── dir2/
    └── file
```
matches
- dir/dir1/file
- dir/dir2/file
- dir/dir1/dir3/file

## *: Same as regex's
`dir/file*` on 
```
dir/
├── file1
├── file2
└── 3file
```
matches
- dir/file1
- dir/file2

## !: exclude files
```
dir
!dir/dir2
```
on
```
dir/
├── dir1/
│   └── file1
└── dir2/
    └── file2
```
matches
- dir/dir1/file1

## #: comment

```
#path/to/file
#!path/to
```
nothing will happen

namui-cli/src/util/get_namui_bundle_list/get_namui_bundle_list.rs
```diff
@@ -0,0 +1,22 @@
+use super::{
```
namse, line 1: I think directory name should be bundle_list, not get_namui_bundle_list..?

`get_namui_bundle_list/get_namui_bundle_list.rs` --> `bundle_list/get_namui_bundle_list`

namui-cli/src/util/get_namui_bundle_list/lexicon/namui_bundle_list.rs
```diff
@@ -0,0 +1,45 @@
+use super::{ExcludeOperation, IncludeOperation};
+use std::{collections::HashMap, path::PathBuf};
+
+type SrcDestPathMap = HashMap<PathBuf, PathBuf>;
+
+#[derive(PartialEq, Eq, Debug)]
+pub struct NamuiBundleList {
```
namse, line 7: I suggest rename it again... to "NamuiBundleMenifest" or something like that. How about this?
It's not actually the list of bundle, it is kind of information what kind of bundle namui-cli should take. So it should have prefer name for the purpose.

namui-cli/src/util/get_namui_bundle_list/lexicon/namui_bundle_list.rs
```diff
@@ -0,0 +1,45 @@
...
+    include: Vec<IncludeOperation>,
+    exclude: Vec<ExcludeOperation>,
+}
+impl NamuiBundleList {
+    pub fn new(include: Vec<IncludeOperation>, exclude: Vec<ExcludeOperation>) -> Self {
+        Self { include, exclude }
+    }
+
+    pub fn flatten(
```
namse, line 16: I think naming is not good... `query`?

namse (changes requested):

namui-cli/src/util/get_namui_bundle_list/lexicon/namui_bundle_list.rs
bigfoodK: Aaah that looks good

namui-cli/src/util/namui_bundle_manifest/get_namui_bundle_manifest.rs
```diff
@@ -0,0 +1,24 @@
...
+    lexicon::{Lexer, NamuiBundleManifest},
+    token::Tokenizer,
+};
+use std::{fs, path::PathBuf};
+
+pub fn get_namui_bundle_manifest(
+    project_root_path: &PathBuf,
+) -> Result<NamuiBundleManifest, String> {
+    let namui_bundle_list_path = project_root_path.join(".namuibundle");
+    let namui_bundle_list_string = match namui_bundle_list_path.exists() {
```
namse, line 11: Here rename too?

namse (changes requested):

namui-cli/src/util/namui_bundle_manifest/get_namui_bundle_manifest.rs
bigfoodK: HAWK EYE

namse (approved):

PR #150, opened 2022-04-13
Build conditionally according to namui-cfg

bigfoodK:
# New cli commands
## `namui target <target_procedure>`
Set working target to <target_procedure>. This will be stored in `namui-cli/namui_user_config.json`.

It will set namui_cfgs name of `target_env`, `target_arch` and `target_platform`
- `target_env`
  - dev
  - release
- `target_arch`
  - wasm
- `target_platform`
  - web
  - electron

<target_procedure> can be
- dev-wasm-web
- release-wasm-web
- dev-wasm-electron
- release-wasm-electron

## `namui print cfg`
Print namui_cfg from `namui_user_config.json` in `key=value` form
ex)
```
target_env=dev
target_arch=wasm
target_platform=web
```

# New namui feature(atrribute)
`#[namui_cfg(<expr>)]`
It can be used the same as std::cfg. ex) `#[namui_cfg(all(target_env = "dev", not(target_platform = "electron")))]`
**but it can only be used with fucntion**. ex) 
```
#[namui_cfg(expr)]
pub fn func() {...}
```

## What happens if build on different  target from user config?
If you run `namui target dev-wasm-electron` and run `namui release-wasm-web`, cli will ignore user config.
Code will be built with cfgs of `release-wasm-web` procedure.


namui-cfg/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+
+static NAMUI_CFG_ENV_PREFIX: &str = "NAMUI_CFG_";
+
+type CfgMap = HashMap<String, String>;
+
+lazy_static! {
+    static ref CFG_VALUES: Vec<CfgValue> = {
+        let mut cfg_map = CfgMap::new();
+        load_cfg_from_namui_cli(&mut cfg_map).unwrap();
+        load_cfg_from_env(&mut cfg_map).unwrap();
```
bigfoodK, line 14: Load cfg from namui-cli first, then overwrite cfg from env.

namui-cfg/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+                true => CfgValue::Name(key.clone()),
+                false => CfgValue::KeyPair(key.clone(), value.clone()),
+            })
+            .collect()
+    };
+}
+
+#[proc_macro_attribute]
+pub fn namui_cfg(attr: TokenStream, item: TokenStream) -> TokenStream {
+    let expression = parse(&format!("cfg({})", attr.to_string().replace("\n", " "))).unwrap();
```
bigfoodK, line 27: If there is newline, parsing error will occur. so remove newline with `attr.to_string().replace("\n", " ")`

.github/workflows/test-namui.yml
```diff
@@ -9,4 +9,4 @@ jobs:
     steps:
       - uses: actions/checkout@v2
 
-      - run: docker build --target test ./namui
+      - run: docker build --target test -f namui/Dockerfile .
```
bigfoodK, line 12: Run docker from `namseent/` not `namseent/namui` to include `namui-cfg`.

namui-cli/src/util/namui_user_config/get_namui_user_config.rs
```diff
@@ -0,0 +1,16 @@
+use super::NamuiUserConfig;
+use crate::util::get_cli_root_path;
+use std::fs;
+
+pub fn get_namui_user_config() -> Result<NamuiUserConfig, Box<dyn std::error::Error>> {
+    let namui_user_config_path = get_cli_root_path().join("namui_user_config.json");
+    let namui_user_config: NamuiUserConfig = match namui_user_config_path.exists() {
+        true => {
+            let file = fs::read(&namui_user_config_path)?;
+            serde_json::from_slice(&file)?
```
namse, line 10: good usage of question mark operator!

.github/workflows/test-namui.yml
namse: it's not a good idea because Dockerfile assume it run on the same directory. 
You need to find the other way to copy cfg file

namse (changes requested):

namse:
I think arch is good but env and platform has some problem.

> - `target_env`
>   - dev
>   - release
> - `target_platform`
>   - web
>   - electron

[`target_env`](https://doc.rust-lang.org/reference/conditional-compilation.html#target_env) is about ABI or platform. `dev` and `release` is [`profile`](https://doc.rust-lang.org/cargo/reference/profiles.html). So, `target_platform` must be `target_env`. 

I think wasm-unknown-unknown broke everything. other system seems ok...

We won't use custom arch. so we don't have to specify arch for namui, but just for rustc.
What we customly use are platform and state-of-building.

# platform

electron, web-browser, nintendo switch.

# state-of-buildling(temp name)

dev and package. hot-reload-on and off? watch? yes, i think it's about watch (-w).


let's think about how to build windows electron package and linux electron package, ios web-view package. for now, it's not available because there are no way to set os for namui-cli.


I gathered my ideas below.

Here are my idea about cli usage.
```
namui target {triple}
namui start [{triple, optional, default to namui target}]
namui build [{triple}]
```

Here are targets in my opinion
- `wasm-pc-windows-electron`
  - rust target: `wasm-unknown-unknown`
- `wasm-unknown-linux-electron`
  - rust target: `wasm-unknown-unknown`
- `wasm-apple-ios-web`
  - rust target: `wasm-unknown-unknown`
- `wasm-unknown-unknown-web`
  - rust target: `wasm-unknown-unknown`
- `x86_64-pc-windows-msvc`
  - rust target: `x86_64-pc-windows-msvc`

these mean, if we try to use custom targets over rust targets, we have to support namui user can specify this configuration using namui-cfg, not rust's cfg, because rust's cfg doesn't have enough information against our namui cfg.

For example, let's think about `wasm-apple-ios-web`. It will run on ios with ios's file system. but rust's cfg(target_os) is unknown. so we have to support namui_target_os = ios.

```
namui start wasm-apple-ios-web

namui target wasm-apple-ios-web
namui build 
```


And `wasm-unknown-unknown-web`, even if the platform is same as web with `wasm-apple-ios-web`, but it doesn't support file system. because os is unknown. so user can notice whether os is unknown of ios in namui environment.

```
namui target wasm-unknown-unknown-web
namui start
namui build 
```

to support hot-reload or watch-reload, we have to let our system know whether the watch option is on or not. let's use environment variable `NAMUI_WATCH_RELOAD`, so electron check `NAMUI_WATCH_RELOAD` to use http url for index.html file instead of file:// url.


So, conclusion! 

Let's change 
1. namui-cli commands
2. namui-targets and namui-cfg options

.github/workflows/test-namui.yml
bigfoodK: How about moving `namui-cfg/` crate down the `namui/` directory?
This is because other crates use re-exported `namui_cfg` through namui instead of calling namui-cfg directly.
In other words, we don't need to call Namui-cfg directly from other crates that use Namui

namse: discussion on slack https://projectluda.slack.com/archives/C02BNFY26NN/p1649925529501659?thread_ts=1649891582.830459&cid=C02BNFY26NN

namse:
Open this if you ready

bigfoodK:
# Changes
## Cli style
### before
`namui dev-unknown-web`
`namui release-unknown-web`
### after
`namui start wasm-unknown-web`
`namui build wasm-unknown-web`

## Targets
- wasm-windows-electron
- wasm-linux-electron
- wasm-unknown-web

## Cfgs
- target_os
  - linux
  - windows
  - unknown
- target_env
  - web
  - electron
- target_arch
  - wasm
- watch_reload

namse (approved):
There are following things to do.
1. Test runner for wasm-linux-electron, wasm-windows-electron in github action
2. (optional) Dockerize electron

I think it can be done at next time.

PR #156, opened 2022-04-18
Use github action cache

namse:
# What I did
1. Restore Cargo.lock for executable --> [reason](https://doc.rust-lang.org/cargo/faq.html#why-do-binaries-have-cargolock-in-version-control-but-not-libraries)
2. Use github action cache for cargo outputs
3. Chmod all docker container outputs

Before Cache
image

After Cache
image


luda-editor/luda-editor-client/.gitignore
```diff
@@ -1,6 +1,16 @@
...
+# Generated by Cargo
+# will have compiled files and executables
+debug/
+target/
+
+# These are backup files generated by rustfmt
 **/*.rs.bk
-Cargo.lock
-bin/
-pkg/
```
bigfoodK, line 5: `wasm-pack` will create `luda-editor-client/pkg`
I think it's better to move `pkg` under `target` dir in future.
But for now, I think pkg should be ignored.

bigfoodK: Ahhh wasm-pack is already doing it. Forget it😊.

bigfoodK (approved):

PR #158, opened 2022-04-19
Use ghcr instead of artifact in github actions

namse:
ghcr is faster than artifact because ghcr uses docker, which has incremental, layer-based image system.

image
Before: 9m
After: 6m


.github/actions/init-test-job/action.yml
```diff
@@ -13,7 +13,7 @@ runs:
       run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
 
     - shell: bash
-      run: rustup target add wasm32-unknown-unknown
+      run: time rustup target add wasm32-unknown-unknown
```
namse, line 16: time is for check elapsed time.

bigfoodK (approved):

PR #165, opened 2022-05-11
Get bounding box of rendering tree

namse:
# Demo
## Text bounding box
image
image
Image from iOS

## Rendering Tree test
image
image

# What I did
- Implement getting bounding box of rendering tree, rendering data, draw command!
- Use bounding box to check mouse is in text or not.

### DIFF
--- a/namui/src/namui/render/rendering_tree.rs
+++ b/namui/src/namui/render/rendering_tree.rs
@@ -379,6 +379,7 @@ impl RenderingTree {
         }
     }
 
+    // NOTE: matrix is reversed to keep xy in same.
     fn is_point_in(&self, xy: &Xy<f32>, matrix: &Matrix3x3) -> bool {
         match self {
             RenderingTree::Children(ref children) => {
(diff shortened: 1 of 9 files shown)


namui/src/namui/render/rendering_tree.rs
```diff
@@ -524,18 +525,167 @@ impl RenderingTree {
...
+                                        let sixteen_xys = xs
+                                            .iter()
+                                            .zip(ys.iter())
+                                            .map(|(x, y)| Xy { x: *x, y: *y });
+
+                                        let difference_area_xys = sixteen_xys.filter(|xy| {
+                                            (clip_bounding_box.is_xy_outside(&xy)
+                                                || clip_bounding_box.is_xy_on_border(&xy))
+                                                && !bounding_box.is_xy_outside(&xy)
+                                        });
```
namse, line 609: Can you get what I did?

namui/src/namui/skia/base/path_builder.rs
```diff
@@ -50,7 +50,7 @@ impl PathBuilder {
     }
     pub fn stroke(&mut self, options: StrokeOptions) -> Result<(), ()> {
         self.commands.push(PathCommand::Stroke(options));
-        Ok(())
+        Ok(()) // TODO: This is false Ok. Make it sure with stroke execution.
```
namse, line 53: I just found here is potential problem, but not critical right now. so just left comment.

namui/src/namui/render/rendering_tree.rs
bigfoodK: At first, there was a bit of confusion because I thought of the LTRB of css. But it's okay now.

image

bigfoodK (approved):

PR #166, opened 2022-05-11
Prebuilt scroll and list view

namse:
Welcome to namui-prebuilt!

namui-prebuilt is prebuilt namui views library!

# What i did
- ScrollView
  - ScrollView shows scroll bar if content size exceed View's size.
  - Currently only vertical scroll is supported.
- ListView
  - ListView is a vertical list view with fixed height items.
  - It uses ScrollView.
  - ListView only draw the item that will be shown, calculating the scroll.

# Scroll View Design
- Scroll bar has been designed like ios's scroll bar.
  - image
  - (please ignore red texts. i just goggled image!)
  - Because we cannot push content inner of scroll because we don't know that when we draw content.

# Demo

https://user-images.githubusercontent.com/3580430/167909205-1dbc24c9-1576-43f5-a0f4-45ec8e2bb7cd.mp4

No frame drop. No performance problem.


namui-prebuilt/src/list_view.rs
```diff
@@ -0,0 +1,90 @@
...
+        let item_len = items_iter.len();
+
+        if item_len == 0 {
+            return RenderingTree::Empty;
+        }
+
+        let scroll_y = self.scroll_view.scroll_y;
+        let visible_item_start_index = (scroll_y / item_wh.height).floor() as usize;
+        let visible_item_end_index = ((scroll_y + height) / item_wh.height).ceil() as usize;
+        let visible_item_count = visible_item_end_index - visible_item_start_index + 1;
```
namse, line 54: ListView only draw the item that will be shown, calculating the scroll.

namui-prebuilt/src/list_view.rs
```diff
@@ -0,0 +1,90 @@
...
+        &self,
+        Props {
+            x,
+            y,
+            height,
+            scroll_bar_width,
+            item_wh,
+        }: Props,
+        items: impl IntoIterator<Item = TItem, IntoIter = TIterator>,
+        item_render: impl Fn(TItem, usize) -> RenderingTree,
```
namse, line 39: I couldn't put them into props because it need to use `impl trait`.

namui-prebuilt/src/scroll_view.rs
```diff
@@ -0,0 +1,139 @@
+use namui::prelude::*;
```
namse, line 1: this code is very same with luda-editor-client's scroll.

namui-prebuilt/src/list_view.rs
```diff
@@ -0,0 +1,113 @@
...
+        Self {
+            scroll_view: scroll_view::ScrollView::new(),
+        }
+    }
+    pub fn update(&mut self, event: &dyn std::any::Any) {
+        self.scroll_view.update(event);
+    }
+    pub fn render<TItem, TIterator, TItems, TItemRender>(
+        &self,
+        props: Props<TItem, TIterator, TItems, TItemRender>,
```
namse, line 38: Note that this props is not reference type unlike other props.
I think we should not use reference type for props because props struct is collection of parameter, and the purpose is named argument which rust doesn't support yet.

So if you want to move value, not reference, you cannot with `&Props` because it's reference type. But you can move value without Props.

To keep it same mood, I suggest not to use `&Props`, but just `Props`.

It needs to be discussion(https://github.com/NamseEnt/namseent/issues/167) so I just remove reference only here.

namui-prebuilt/src/list_view.rs
```diff
@@ -0,0 +1,113 @@
...
+        height: 100.0,
+        scroll_bar_width: 10.0,
+        item_wh: Wh::new(100.0, 100.0),
+        items: &items,
+        item_render: |_item, _index| namui::render![],
+    };
+
+    let list_view = ListView::new();
+    list_view.render(props);
+}
```
bigfoodK, line 113: Is it intended to leave this function?

namui-prebuilt/src/list_view.rs
bigfoodK: I think it's okay.
Why don't we apply it to scroll_view too?

namui-prebuilt/src/list_view.rs
```diff
@@ -0,0 +1,113 @@
...
+        let visible_item_end_index =
+            ((scroll_y + props.height) / props.item_wh.height).ceil() as usize;
+        let visible_item_count = visible_item_end_index - visible_item_start_index + 1;
+
+        let visible_items = items_iter
+            .enumerate()
+            .skip(visible_item_start_index)
+            .take(visible_item_count);
+
+        let rendered_items = visible_items.map(|(index, item)| {
```
bigfoodK, line 63: Now it provides index of just visible items, not of all items.
Do we need index of visible items? Can you give me an example of when to use the index?

namui-prebuilt/src/list_view.rs
namse: Yes, I wanted to test whether `items` and `item_render` type inference work or not in this package.

namui-prebuilt/src/list_view.rs
namse: Actually I'm not sure about it.
You know, in javascript, they give us index in `forEach` and `map`.

In rust, maybe they can use enumerate to get index of item if they really want.

```
items: items.iter().enumerate(),
```
it works with current code. hmm. Ok. I will change it as you suggest!

namui-prebuilt/src/list_view.rs
namse: Let me think about this problem more, then I will replace all later if I'm sure about this.

bigfoodK (approved):

PR #175, opened 2022-05-18
Fallback emoji font

namse:
Yes, namui support emoji using [NotoColorEmoji](https://github.com/googlefonts/noto-emoji)!

image

# How it works
1. namui try to draw text with font that user set when they call namui::text.
2. If namui fail to draw text with font, it use 'fallback font'.
3. Ah, Yas, I make namui load noto emoji as fallback font.

### DIFF
--- a/namui/src/namui/draw/text.rs
+++ b/namui/src/namui/draw/text.rs
@@ -1,36 +1,135 @@
-use std::borrow::Borrow;
-
-use crate::namui::{self, skia::*, NamuiContext};
-
 use super::{TextAlign, TextBaseline, TextDrawCommand};
+use crate::namui::{skia::*, NamuiContext};
+use std::{
+    collections::{HashMap, VecDeque},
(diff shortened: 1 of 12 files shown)


namui/src/namui/draw/text.rs
```diff
@@ -1,36 +1,135 @@
...
+            continue;
+        }
+
+        let available_glyph_id_and_index_and_width: Vec<(u16, usize, f32)> = {
+            let available_glyph_ids: Vec<_> = available_glyph_id_and_indexes
+                .iter()
+                .map(|(glyph_id, _)| *glyph_id)
+                .collect();
+
+            let widths = font.get_glyph_widths(&available_glyph_ids, Option::Some(paint));
```
namse, line 105: It's because I want to call external call(`font.get_glyph_widths`) only once per font.

bigfoodK (approved):

PR #182, opened 2022-05-23
Add namui-user-config lib

bigfoodK:
# Why namui-user-config?
An error occurred when namui-cfg failed to run namui-cli while running namui test.
Because there was no process to install namui-cli on the test host.
But this is strange. So instead of installing namui-cli on test host, I decided to create a library that helps namui-cfg and namui-cli manage user configs.
And this is it.

# Where to save user config
It was previously stored in the root directory of namui-cli.
For now, save it to a more general user settings directory.
Please see below


namui-user-config/src/get_user_config_path.rs
```diff
@@ -0,0 +1,18 @@
...
+impl Display for GetUserConfigPathError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Failed to get project directory")
+    }
+}
+impl Error for GetUserConfigPathError {}
+
+pub fn get_user_config_path() -> Result<PathBuf, GetUserConfigPathError> {
+    match ProjectDirs::from("com", "namseent", "namui") {
+        Some(project_dir) => Ok(project_dir.config_dir().join("namui_user_config.json")),
```
bigfoodK, line 15: In my case(wsl ubuntu): `/home/bigfood/.config/namui/namui_user_config.json`
FYI: [directories](https://crates.io/crates/directories)

bigfoodK:
> Why not just combine it into namui-cfg?

@namse 
This is because proc-macro crate can only export proc-macro.
image

namui-user-config/Cargo.lock
```diff
@@ -0,0 +1,207 @@
+# This file is automatically @generated by Cargo.
```
namse, line 1: This should not added in commit because namui-user-config is just lib. please remove it

namui-user-config/src/set_user_config.rs
```diff
@@ -33,11 +33,15 @@ pub fn set_namui_user_config(target: &Target) -> Result<(), Box<dyn std::error::
...
     )?;
-
-    println!("Settings have been saved.");
     Ok(())
 }
+
+fn ensure_user_config_dir(user_config_path: &PathBuf) -> Result<(), std::io::Error> {
+    let mut user_config_dir = user_config_path.clone();
+    user_config_dir.pop();
+    create_dir_all(user_config_dir)
```
namse, line 46: What's going in this code? why pop? to get directory path, not file path?

namse (changes requested):

namui-user-config/Cargo.lock
bigfoodK: Ah!

namui-user-config/src/set_user_config.rs
bigfoodK: Yes, ensuring user_config.json located directory

namse (approved):

PR #184, opened 2022-05-23
Protect managers with reentrant mutex

namse:
# What I did
When I develop text input, I faced with recursive lock problem. I locked mutex again.
You know, recursive locking is very usually used in C# or other language(I guess..!?). So I put reentrant mutex for managers.


namui/src/namui/manager/mod.rs
```diff
@@ -16,10 +17,24 @@ pub mod mock;
...
+    MANAGERS.get().expect("managers not initialized").lock()
+}
+
+pub(crate) fn set_managers(managers: Managers) {
+    if MANAGERS
+        .set(ReentrantMutex::new(Arc::new(managers)))
+        .is_err()
+    {
+        panic!("managers already initialized");
+    }
```
namse, line 39: This is the core of this PR.
1. Use Reentrant Mutex for `fn managers()`
2. Remove all Boxing of Managers inside
3. manager's pub fn should use immutable reference

namui/src/namui/manager/web/image_manager.rs
```diff
@@ -16,13 +16,13 @@ pub struct ImageManager {
...
-        Arc::new(Self {
+    pub fn new() -> Self {
+        Self {
             image_map: DashMap::new(),
             image_requested_set: Mutex::new(HashSet::new()),
-        })
+        }
     }
-    pub fn try_load(self: Arc<Self>, url: &String) -> Option<Arc<Image>> {
+    pub fn try_load(&self, url: &String) -> Option<Arc<Image>> {
```
namse, line 25: > manager's pub fn should use immutable reference

namui/src/namui/manager/common/typeface_manager.rs
```diff
@@ -1,26 +1,25 @@
+use dashmap::DashMap;
+
 use crate::namui::{self, Typeface, TypefaceType};
-use std::collections::HashMap;
 
 pub struct TypefaceManager {
-    typefaces: HashMap<TypefaceType, Box<Typeface>>,
+    typefaces: DashMap<TypefaceType, Typeface>,
```
namse, line 6: To support immutable reference with manager, I replaced HashMap with DashMap which support lock-free map with immutable reference.
If you want to put String inside of manager struct, you can use Mutex.
```rs
pub struct YourManager {
    id: Mutex<String> // << like this
}
```

bigfoodK (approved):

PR #189, opened 2022-05-24
Check Cargo.lock in CI

namse:
I found that sometime we don't push updated Cargo.lock when we update Cargo.toml. I found how CI can catch it before merge pull requests.

context: https://github.com/rust-lang/rust-analyzer/issues/12219

### DIFF
--- a/.github/actions/namui-test/action.yml
+++ b/.github/actions/namui-test/action.yml
@@ -20,13 +20,12 @@ runs:
       shell: bash
       run: time namui test ${{ inputs.target }} -m ${{ inputs.path }}/Cargo.toml
 
-    - name: Check up to date Cargo.lock checked
-      uses: technote-space/get-diff-action@v6
-      with:
-        FILES: |
-          Cargo.lock
-
-    - name: Confirm Cargo.lock
-      if: env.GIT_DIFF
+    - name: Did you push Cargo.lock
       shell: bash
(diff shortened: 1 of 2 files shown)


namse:
Test will be failed because I didn't push updated Cargo.lock in luda-editor-client. I'm testing it and will update Cargo.lock after test.

lnicola:
The idea was to run `cargo check --frozen` or some equivalent of it. If you updated `Cargo.toml`, but did not commit `Cargo.lock`, `cargo check --frozen` will fail because it wants to update the lockfile but is not allowed to.

namse:
@lnicola But, as you said, isn't it make diff if comitter didn't push updated Cargo.lock? I just thought like that and I didn't want to change my system to run with --frozen args.

namse:
(And really thank you to review this PR!!!!!! ^^)

lnicola:
If you add a feature, change the crate description, or replace `1.0` with `1.0.0` in `Cargo.toml`, you don't need to commit the lockfile. And the test in this PR only runs `git diff`, without even looking at the file names, so I don't really understand it.

lnicola:
Yeah, you're right, I was thinking of `git show`. It seems a bit round-about, but I suppose it could work.

namse:
Yes I agree it's round-about haha

bigfoodK (approved):

PR #193, opened 2022-05-25
Hide text input focus information

namse:
I removed focus boolean from text_input because it would not sync if user didn't update it.

Only single text_input can be focused, so I decide hiding focus information from text_input, and put it into manager.

During that editing, I passed text_input id in blur event so user can check that text input is done or not.

### DIFF
--- a/namui/sample/text-input/src/lib.rs
+++ b/namui/sample/text-input/src/lib.rs
@@ -15,6 +15,7 @@ struct TextInputExample {
     left_text: String,
     center_text: String,
     right_text: String,
+    left_value: Option<f32>,
 }
 
 impl TextInputExample {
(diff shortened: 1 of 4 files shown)


namui/sample/text-input/src/lib.rs
```diff
@@ -145,12 +168,18 @@ impl Entity for TextInputExample {
                 text_input::Event::TextUpdated(text_updated) => {
                     if self.left_text_input.get_id().eq(&text_updated.id) {
                         self.left_text = text_updated.text.clone();
+                        self.left_value = self.left_text.parse().ok(); // NOTE: You don't have to check value in here, it's would be better UX checking it on blur.
```
bigfoodK, line 171: 👍

bigfoodK (approved):

PR #196, opened 2022-05-30
Fix text input mouse click x locality

namse:
I found that `text input selection changing by click` doesn't work. It's because of mouse x was global, but expected local x. I fixed it.


namse:
There are some non-relevant codes in this commit. I think it's ok to merge together.

bigfoodK (approved):

PR #209, opened 2022-06-13
Scale on render

namse:
image

scale? zoom? 

scale(x, y)
zoom(x, y)

which one do you prefer?

if zoom good -> will change name before merge

### DIFF
--- a/namui/src/namui/render/rendering_tree/mod.rs
+++ b/namui/src/namui/render/rendering_tree/mod.rs
@@ -27,6 +27,7 @@ impl SpecialRenderingNode {
             SpecialRenderingNode::Absolute(node) => &node.rendering_tree,
             SpecialRenderingNode::Rotate(node) => &node.rendering_tree,
             SpecialRenderingNode::Custom(node) => &node.rendering_tree,
+            SpecialRenderingNode::Scale(node) => &node.rendering_tree,
         }
     }
 }
@@ -97,6 +98,15 @@ impl RenderingTree {
                     rotate.rendering_tree.draw(namui_context);
 
                     canvas.rotate(-rotate.ccw_radian);
+                }
+                SpecialRenderingNode::Scale(scale) => {
+                    let canvas = namui_context.surface.canvas();
+
+                    canvas.scale(scale.x, scale.y);
+
+                    scale.rendering_tree.draw(namui_context);
+
+                    canvas.scale(1.0 / scale.x, 1.0 / scale.y);
                 }
                 _ => {
                     special.get_rendering_tree().draw(namui_context);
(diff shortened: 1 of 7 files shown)


bigfoodK:
Scale seem to be better. I think it is because I often used it on css and canvas.

namui/src/namui/render/rendering_tree/mod.rs
```diff
@@ -98,6 +99,15 @@ impl RenderingTree {
...
                     canvas.rotate(-rotate.ccw_radian);
                 }
+                SpecialRenderingNode::Scale(scale) => {
+                    let canvas = namui_context.surface.canvas();
+
+                    canvas.scale(scale.x, scale.y);
+
+                    scale.rendering_tree.draw(namui_context);
+
+                    canvas.scale(1.0 / scale.x, 1.0 / scale.y);
```
bigfoodK, line 109: Is it `Divided by zero` safe?

namui/src/namui/render/rendering_tree/visit.rs
```diff
@@ -244,6 +252,12 @@ impl RenderingTree {
                             height: bottom - top,
                         };
                     }
+                    SpecialRenderingNode::Scale(scale) => {
+                        bounding_box.x *= scale.x;
+                        bounding_box.y *= scale.y;
+                        bounding_box.width *= scale.x;
+                        bounding_box.height *= scale.y;
+                    }
```
bigfoodK, line 260: Is it okay to do it after rotate it?
In canvas, axis will rotate together on rotating
image

But this code looks like this
image

Is this the difference between Namui and Canvas?

namse: As you see [line 210](https://github.com/NamseEnt/namseent/pull/209/files/e250b8bd8461828b4e63c3877080ea00001edf67#diff-2dce2d7bc1e94dbad3db40530d42a7112141839460fb4e29a1e21f407a2a3e57R210), when it rotates bounding box, it just expands.

like this.
image

Because it's just 'bounding box', not 'bounding path' so I just did it like this.

bigfoodK: Yes, bounding box will be scaled with fixed axis.
But actually being drawn on canvas will be scaled with rotated axis. Isn't it?

namse: The current bounding box of loop is the being drawn on canvas. So it already drawn in rotated axis.

I mean, let's thing about the tree (parent) 1-2-3-4-5 (children).
if 5 is target, 1,2,3,4 is ancestors. I reversed it's order at line [122](https://github.com/NamseEnt/namseent/pull/209/files/e250b8bd8461828b4e63c3877080ea00001edf67#diff-2dce2d7bc1e94dbad3db40530d42a7112141839460fb4e29a1e21f407a2a3e57R122) so it become 4,3,2,1.

The target, 5, has bounding box already. So this code will process with that bounding box.
if 4 is rotate, 4 will rotate 5's bounding box. ok.
then if 3 is scale, it scale rotated bounding box. ok.

you got it?

bigfoodK: Aaaaah, got it.

bigfoodK (approved):

PR #214, opened 2022-06-13
Use Url for dirent and image source

namse:
I found that it's very difficult to set image source from bundle. So I unified bundle dirent and previous string based url into servo's url crate.


bigfoodK (approved):

bigfoodK:
@namse Is it okay to merge

PR #217, opened 2022-06-14
Cache font information for optimization

namse:
When I render so many text, it makes lags so I optimized rendering and drawing text with cache.

before
image

after
image

(I reduced event count so comparison is not accurate.)

### DIFF
--- a/namui/src/namui/skia/base/paint_builder.rs
+++ b/namui/src/namui/skia/base/paint_builder.rs
@@ -1,6 +1,4 @@
-use crate::{
-    namui::skia::StrokeJoin, BlendMode, Color, ColorFilter, Paint, PaintStyle, StrokeCap, Xy,
-};
+use crate::{namui::skia::StrokeJoin, BlendMode, Color, ColorFilter, Paint, PaintStyle, StrokeCap};
 use once_cell::sync::OnceCell;
 use ordered_float::OrderedFloat;
 use serde::Serialize;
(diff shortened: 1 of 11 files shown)


namui/src/namui/skia/base/paint_builder.rs
```diff
@@ -84,30 +82,19 @@ impl PaintBuilder {
...
+            self.color,
+            self.paint_style.as_ref(),
+            self.anti_alias,
+            self.stroke_width,
+            self.stroke_cap.as_ref(),
+            self.stroke_join.as_ref(),
+            self.color_filter
+                .as_ref()
+                .map(|(color, blend_mode)| Self::get_or_create_color_filter(&color, &blend_mode)),
+        );
```
namse, line 97: This is for making paint immutable. I removed all set functions.

namui/src/namui/skia/base/paint_builder.rs
```diff
@@ -126,6 +113,19 @@ impl PaintBuilder {
...
+            &self.color,
+            &self.paint_style,
+            &self.anti_alias,
+            &self.stroke_width,
+            &self.stroke_cap,
+            &self.color_filter,
+        ))
+        .unwrap()
+        .into()
+    }
```
namse, line 128: used for cache hash

bigfoodK (approved):

PR #218, opened 2022-06-19
Fix get_pressing_buttons to return right buttons

bigfoodK:
# What happened
When the wheel button was pressed, get_pressing_buttons() acted as if the right button was pressed.

### DIFF
--- a/namui/src/namui/manager/web/mouse_manager.rs
+++ b/namui/src/namui/manager/web/mouse_manager.rs
@@ -149,8 +149,8 @@ fn get_pressing_buttons(mouse_event: &web_sys::MouseEvent) -> HashSet<crate::Mou
 
     const MOUSE_BUTTONS_CONVERTING_TUPLES: [(u16, crate::MouseButton); 3] = [
         (1 << 0, crate::MouseButton::Left),
-        (1 << 1, crate::MouseButton::Middle),
-        (1 << 2, crate::MouseButton::Right),
+        (1 << 1, crate::MouseButton::Right),
+        (1 << 2, crate::MouseButton::Middle),
     ];
 
     HashSet::from_iter(


namui/src/namui/manager/web/mouse_manager.rs
```diff
@@ -149,8 +149,8 @@ fn get_pressing_buttons(mouse_event: &web_sys::MouseEvent) -> HashSet<crate::Mou
 
     const MOUSE_BUTTONS_CONVERTING_TUPLES: [(u16, crate::MouseButton); 3] = [
         (1 << 0, crate::MouseButton::Left),
-        (1 << 1, crate::MouseButton::Middle),
-        (1 << 2, crate::MouseButton::Right),
+        (1 << 1, crate::MouseButton::Right),
+        (1 << 2, crate::MouseButton::Middle),
     ];
```
bigfoodK, line 154: > 0: Main button pressed, usually the left button or the un-initialized state
> 1: Auxiliary button pressed, usually the wheel button or the middle button (if present)
> 2: Secondary button pressed, usually the right button
> 3: Fourth button, typically the Browser Back button
> 4: Fifth button, typically the Browser Forward button
[mdn](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button#value)

namse: Isn't it right by mdn doc?

namse (changes requested):

namui/src/namui/manager/web/mouse_manager.rs
namse: You put wrong information.
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons
image

So it's right

namse (approved):

PR #220, opened 2022-06-20
Cache image info fixing wrong info processing

namse:
I found that it's often called to get image info from canvas kit but it wasn't cached. I put cache and during that fixing, I found the bug of my previous code about image info processing. I fixed it too.

### DIFF
--- a/namui/src/namui/skia/web/image.rs
+++ b/namui/src/namui/skia/web/image.rs
@@ -11,62 +11,56 @@ unsafe impl Send for CanvasKitImage {}
 pub struct Image {
     id: String,
     #[serde(skip)]
-    pub canvas_kit_image: CanvasKitImage,
+    pub(crate) canvas_kit_image: CanvasKitImage,
+    image_info: PartialImageInfo,
 }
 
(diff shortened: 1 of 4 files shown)


namui/src/namui/skia/web/image.rs
```diff
@@ -11,62 +11,56 @@ unsafe impl Send for CanvasKitImage {}
...
-                rgb565 => ColorType::Rgb565,
-                rgba8888 => ColorType::Rgba8888,
-                bgra8888 => ColorType::Bgra8888,
-                rgba1010102 => ColorType::Rgba1010102,
-                rgb101010x => ColorType::Rgb101010x,
-                gray8 => ColorType::Gray8,
-                rgbaf16 => ColorType::RgbaF16,
-                rgbaf32 => ColorType::RgbaF32,
-            },
-        }
```
namse, line 63: this code doesn't work but i didn't know that :(

namse:
Actually PR test was failed. I think it's the bug of github.
image

namse:
I made a github support ticket about it.
https://support.github.com/ticket/personal/0/1674958

Anyway don't worry about it, the test on master branch was passed.

PR #222, opened 2022-06-20
Fix local_xy failure with scale

namse:
I found that local_xy doesn't work with scale. I fixed it.
Live tested with namui-animation-editor(in dev)

### DIFF
--- a/namui/src/namui/manager/web/text_input_manager.rs
+++ b/namui/src/namui/manager/web/text_input_manager.rs
@@ -268,7 +268,7 @@ impl TextInputManager {
                     render::SpecialRenderingNode::Custom(custom) => {
                         if let Some(custom_data) = custom.data.downcast_ref::<TextInputCustomData>()
                         {
-                            let is_custom_in_mouse = utils.is_xy_in(&raw_mouse_event.xy);
+                            let is_custom_in_mouse = utils.is_xy_in(raw_mouse_event.xy);
 
                             if is_custom_in_mouse {
                                 return_value = Some(custom_data.clone());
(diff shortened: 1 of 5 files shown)


namui/src/namui/manager/web/text_input_manager.rs
```diff
@@ -268,7 +268,7 @@ impl TextInputManager {
                     render::SpecialRenderingNode::Custom(custom) => {
                         if let Some(custom_data) = custom.data.downcast_ref::<TextInputCustomData>()
                         {
-                            let is_custom_in_mouse = utils.is_xy_in(&raw_mouse_event.xy);
+                            let is_custom_in_mouse = utils.is_xy_in(raw_mouse_event.xy);
```
namse, line 271: Some file edited for removing & of Xy<f32> in parameter

namui/src/namui/render/rendering_tree/visit.rs
```diff
@@ -350,84 +347,173 @@ mod tests {
...
+                            assert_approx_eq!(f32, local_xy.y, -115.0, ulps = 2);
+                            call_count += 1;
+                        }
+                        _ => {}
+                    }
+                }
+            }
+            ControlFlow::Continue(())
+        });
+        assert_eq!(call_count, 11);
```
namse, line 430: Actually the reason why the test on previous PR was passed is, I didn't check 9 and 10. So I made it sure that call_count is 11, so 9 and 10 is called.

bigfoodK (approved):
I saw the alarm late. Sorry!

PR #230, opened 2022-06-24
Make attach_event builder can break chain

namse:
# What's the problem

Can't move same variable to multiple event handler.

image

image

# How to solve that?

## Idea 1. Break chain

image

It can be solved if I break chain, but look at (1). it's too ugly. We have to create new builder because previous builder has been moved.

## Idea 2. use &mut instead move

image

It can break chain, but don't have to re-assign builder because we didn't move builder. Still can chain events.

I believe Idea 2 is the answer. So I impl that.

It's the breaking change because this new attach fn doesn't return builder. So If review pass, I will change every current our code into this style.

- [x] Review Passed
- [x] Change all attach_event
- [x] Last Review

### DIFF
--- a/namui/src/namui/render/special/attach_event.rs
+++ b/namui/src/namui/render/special/attach_event.rs
@@ -129,7 +129,7 @@ impl AttachEventBuilder {
         self.on_wheel = Some(Arc::new(on_wheel));
         self
     }
-    
+
     pub fn on_key_down(&mut self, on_key_down: impl Fn(&KeyboardEvent) + 'static) -> &mut Self {
         self.on_key_down = Some(Arc::new(on_key_down));
         self
--- a/namui/src/namui/render/special/attach_event.rs
+++ b/namui/src/namui/render/special/attach_event.rs
@@ -76,26 +76,15 @@ pub struct AttachEventBuilder {
     pub(crate) on_key_up: Option<KeyboardEventCallback>,
 }
 
-#[derive(Default)]
-pub struct AttachEventBuilder2 {
-    pub(crate) on_mouse_move_in: Option<MouseEventCallback>,
-    pub(crate) on_mouse_move_out: Option<MouseEventCallback>,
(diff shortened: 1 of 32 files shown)


bigfoodK:
Idea 2 is look good to me too.

namse:
Good. then I will make idea 2 as default and change all our code with that.

namse:
@bigfoodK Final review please

namui/src/namui/render/special/attach_event.rs
```diff
@@ -99,38 +99,34 @@ impl RenderingTree {
...
         self.on_mouse_up = Some(Arc::new(on_mouse_up));
         self
     }
 
-    pub fn on_wheel(mut self, on_wheel: impl Fn(&WheelEvent) + 'static) -> Self {
+    pub fn on_wheel(&mut self, on_wheel: impl Fn(&WheelEvent) + 'static) -> &mut Self {
         self.on_wheel = Some(Arc::new(on_wheel));
         self
     }
-
```
bigfoodK, line 126: I think this should also be included.

namui/src/namui/render/special/attach_event.rs
```diff
@@ -99,38 +99,34 @@ impl RenderingTree {
...
-    pub fn on_key_down(mut self, on_key_down: impl Fn(&KeyboardEvent) + 'static) -> Self {
-        self.on_key_down = Some(Arc::new(on_key_down));
-        self
-    }
-
-    pub fn on_key_up(mut self, on_key_up: impl Fn(&KeyboardEvent) + 'static) -> Self {
-        self.on_key_up = Some(Arc::new(on_key_up));
-        self
-    }
```
bigfoodK, line 135: I think this should also be included.

namse: Good Eagle Eye

bigfoodK (approved):

PR #233, opened 2022-06-24
Clean code of namui context

namse:
I moved namui context codes into namui_context directory.

Will open after merging https://github.com/NamseEnt/namseent/pull/228


bigfoodK (approved):
Code became even sexier 😎.

PR #236, opened 2022-06-25
Set renderingTree as mouse cursor

namse:
I was surprised because it was too easy to make. Tested and worked well.

image

That triangle.

### DIFF
--- a/namui/src/namui/manager/web/mouse_manager.rs
+++ b/namui/src/namui/manager/web/mouse_manager.rs
@@ -120,7 +120,7 @@ impl MouseManager {
 
         mouse_manager
     }
-    pub fn set_mouse_cursor(&self, cursor: MouseCursor) {
+    pub fn set_mouse_cursor(&self, cursor: &MouseCursor) {
         let element = namui::window().document().unwrap().body().unwrap();
         element
             .style()
@@ -140,6 +140,7 @@ impl MouseCursor {
             Self::Grab => "grab",
             Self::Move => "move",
             Self::Pointer => "pointer",
+            MouseCursor::Custom(_) => "none",
         }
     }
 }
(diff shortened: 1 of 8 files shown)


namui/src/namui/namui_context/main_loop/mod.rs
```diff
@@ -8,11 +8,9 @@ impl NamuiContext {
         state: &mut dyn Entity<Props = TProps>,
         props: &TProps,
     ) {
-        let mut event_count = 0;
-
         loop {
             let event = self.event_receiver.recv().await.unwrap();
-            event_count += 1;
+            self.event_count += 1;
```
namse, line 13: I forgot to fix it last time

namui/src/namui/render/special/mouse_cursor.rs
```diff
@@ -13,19 +13,20 @@ pub enum MouseCursor {
...
     Move,
     Pointer,
+    Custom(RenderingTree),
 }
 
 #[derive(Serialize, Clone, Debug)]
 pub struct MouseCursorNode {
     pub(crate) rendering_tree: Box<RenderingTree>,
-    pub cursor: MouseCursor,
+    pub cursor: Box<MouseCursor>,
```
namse, line 22: RenderingTree --> MouseCursorNode --> MouseCursor --> Custom --> RenderingTree :: X. Circular!
So I put box here.

bigfoodK:
How about applying css's none cursor when it is a custom cursor?
Then, we only can see custom cursor without css's default cursor.

https://developer.mozilla.org/ko/docs/Web/CSS/cursor

namui/src/namui/manager/web/mouse_manager.rs
```diff
@@ -140,6 +140,7 @@ impl MouseCursor {
             Self::Grab => "grab",
             Self::Move => "move",
             Self::Pointer => "pointer",
+            MouseCursor::Custom(_) => "none",
```
namse, line 143: @bigfoodK I already did it
https://github.com/NamseEnt/namseent/pull/236#issuecomment-1166344180

bigfoodK: Aaaah. sorry

bigfoodK (approved):

PR #237, opened 2022-06-26
Fix rotation angle in cw direction

namse:
I found that I calculated rotation in wrong way. fixed it.

### DIFF
--- a/namui/src/namui/render/rendering_tree/visit.rs
+++ b/namui/src/namui/render/rendering_tree/visit.rs
@@ -79,7 +79,7 @@ impl RenderingTree {
                         xy.y -= absolute.y;
                     }
                     SpecialRenderingNode::Rotate(rotate) => {
-                        xy = rotate.get_matrix().transform_xy(xy);
+                        xy = rotate.get_counter_wise_matrix().transform_xy(xy);
                     }
                     SpecialRenderingNode::Scale(scale) => {
                         xy.x /= scale.x;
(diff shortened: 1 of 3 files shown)


namui/src/namui/render/rendering_tree/visit.rs
```diff
@@ -79,7 +79,7 @@ impl RenderingTree {
                         xy.y -= absolute.y;
                     }
                     SpecialRenderingNode::Rotate(rotate) => {
-                        xy = rotate.get_counter_wise_matrix().transform_xy(xy);
+                        xy = rotate.get_matrix().transform_xy(xy);
```
namse, line 82: rotate = ccw(counter-clock-wise). so it could be cw(clock-wise).

Look at the other code, they minus or divide values. It mean this code should calculate in reversed way.

bigfoodK: Was it ccw? I thought it was cw until now. Because in JavaScript, it's cw.

bigfoodK (dismissed):

namui/src/namui/render/rendering_tree/visit.rs
namse: UH!! WAIT

namse: You are right, why I thought it is ccw? shit, wait please

bigfoodK (approved):

PR #238, opened 2022-06-26
Save Images into zip file

bigfoodK:
# Why do we need this?
Before this, we download multiple image file at the smae time. chrome can allow this, but safari can't.
This pr will make it possible by saving images into one zip file.


image-cropper/src/app/cropper/save_image/get_png_encoded_u8_from_canvas.rs
```diff
@@ -0,0 +1,32 @@
...
+            }) as Box<dyn FnOnce(Blob)>)
+            .into_js_value()
+            .as_ref()
+            .unchecked_ref(),
+            "image/png",
+        )
+        .expect("failed to create blob url from canvas");
+    receiver
+        .await
+        .expect("canvas_to_png_encoded_u8 did not receive data")
```
bigfoodK, line 31: Using callback style function as async style using oneshot.
I don't know if this isn't safe.

image-cropper/src/app/cropper/save_image/save_image.rs
```diff
@@ -0,0 +1,169 @@
...
+    if let Some(first_point) = polygon.first() {
+        context.begin_path();
+        context.move_to(first_point.x as f64, first_point.y as f64);
+        for point in polygon.iter().skip(1) {
+            context.line_to(point.x as f64, point.y as f64);
+        }
+        context.close_path();
+        context.clip();
+    }
+}
```
namse, line 169: Good code

namse (approved):

image-cropper/src/app/cropper/save_image/get_png_encoded_u8_from_canvas.rs
namse: It's ok because they just call the function and don't get the response.

bigfoodK: In fact, u8_array is returned. Does this matter?

namse: > u8_array is returned
No, they don't return in callback. they send via sender. So it's ok.

namse: I mean, javascript will just call this wasm function and that's all. so it's ok that they make our async function starts

bigfoodK: Ah, got it

PR #239, opened 2022-06-27
Fix wrong absolute xy in check

namse:
# Why I didn't know that
the test didn't check how absolute works


namui/src/namui/render/rendering_tree/visit.rs
```diff
@@ -65,32 +65,32 @@ impl RenderingTree {
...
-                        xy.y /= scale.y;
+                        result_xy.x /= scale.x;
+                        result_xy.y /= scale.y;
                     }
                     _ => {}
                 },
                 _ => {}
             }
         }
-        xy
```
namse, line 93: name `xy` was same with parameter `xy`

bigfoodK (approved):

PR #242, opened 2022-06-28
Add zoom tool

bigfoodK:
Add zoom tool.
Shortcut is 4 and Z.
Drag to right bottom will zoom in, left up will zoom out.


namse (approved):
I think you call clone too much :)

PR #250, opened 2022-06-29
Bye bye warning in namui!

namse:
I removed all warning in namui.

`cargo fix --allow-dirty --broken-code` is god


namui/src/lib.rs
```diff
@@ -1,7 +1,10 @@
+#![deny(warnings)]
```
namse, line 1: https://github.com/rust-unofficial/patterns/blob/main/anti_patterns/deny-warnings.md

namui/src/namui/skia/web/paint.rs
```diff
@@ -1,5 +1,5 @@
 use super::*;
-use crate::namui;
+
```
bigfoodK, line 2:
```suggestion
```
I think it would be better to delete the blank line between imports. For the lint to work

namse: Fixed in d48d9fd574bd10dfc30f5a1117e2490107277c4dc

PR #251, opened 2022-06-29
Say hello to new namui types!

namse:
I add some types. You will find how to use it in animation branch PR.


namse:
I'm thinking about unitless, dimentionless types.

namse:
I'm not sure about type system so I won't open it before making sure.

PR #255, opened 2022-07-05
Mock time on test

namse:
I failed the test that uses time system. I forgot mocking time system to run this test.

image

So I mocked time system.

### DIFF
--- a/namui/src/namui/common/set_timeout.rs
+++ b/namui/src/namui/common/set_timeout.rs
@@ -102,35 +102,37 @@ mod tests {
     #[test]
     #[wasm_bindgen_test::wasm_bindgen_test]
     fn pull_timeout_should_sort_callback_by_time() {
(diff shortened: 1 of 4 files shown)


namui/src/namui/common/set_timeout.rs
```diff
@@ -102,35 +102,37 @@ mod tests {
...
-        set_timeout(
-            move || {
-                vec3.lock().unwrap().push(3);
-            },
-            Duration::from_millis(3),
-        );
-
-        while let Some(callback) = pull_timeout(crate::now() + Duration::from_millis(4)) {
-            callback();
+        for _ in 0..1000 {
```
namse, line 105: It's intended, it's a kind of timing problem of `crate::now()` so I make it run multiple times to make a error on non-mocked environment.

bigfoodK (approved):

PR #259, opened 2022-07-06
Support mouse down/up in/out

namse:
I tried to check whether user click outside of object.

I could handle it with background's click event handler, but then click event occur twice- 1) by object, 2) by background.

So I made mouse_down_out and mouse_up_out.


namse:
Usage Example
image

bigfoodK (approved):

PR #266, opened 2022-07-08
Remove num::Primitive from type macro, but only i32 and f32

namse:
1. I found it's not a good practice to use num::ToPrimitive or num::FromPrimitive.
2. rustc fail to infer integer to type because we implemented all integer types so I removed it but only i32. It's not a big deal because all type which use this type macro has f32, so it should be ok.


namse:
Related slack message : https://projectluda.slack.com/archives/C015ELKQE67/p1657303458320849

namse:
I need to think about it more, for the cases of usize.

namse:
usize problem solved. Strangely usize doesn't affect to operation implementation for integer literal.

The important thing about usize operation implementation is because we use i33 and usize very much. 
- i32(for [integer literal fallback ](https://rust-lang.github.io/rfcs/0212-restore-int-fallback.html))
- usize(iterator.count() or [].len(), etc)

bigfoodK (approved):

PR #272, opened 2022-07-18
Draw on top

namse:
`on_top` ignores clipping and draw in last order of drawing stage to draw on most top of display.

Previous :

https://user-images.githubusercontent.com/3580430/179544928-b1bc756e-ef8f-44ca-b25e-a57e3d81fdb8.mp4

After:

https://user-images.githubusercontent.com/3580430/179563138-62c9dc77-67f6-404d-a2dc-d8b32ab33f87.mp4

 What I did
- add `on_top`
- add `stop_propagation` on mouse event


namse:
OK I solved the problem with option 2 and it works very well.

bigfoodK (approved):

PR #273, opened 2022-07-19
Add animation preset squash and stretch and some namui chagnes

namse:
# What I did
- React-ished dropdown
- Add animation preset 'squash and stretch'
- Removed graph editor
- Use matrix instead of xywh for animation image 
- and some small patches

PR #274, opened 2022-07-19
Check namui-animation-editor on namui changes

namse:
I found that namui-animation-editor wasn't tested on master branch when I change namui. it should test editor, and every package which uses namui.


namse:
https://github.com/NamseEnt/namseent/pull/273 will solve the ci error.

bigfoodK (approved):

PR #276, opened 2022-07-20
Run post-test always prevent skipping

namse:
> If a job in a workflow is skipped due to a conditional, it will report its status as "Success". [document](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks)

So I prevent skipping `post-test` and always run.

`+ I removed clean-up because we don't use it now.`


namse:
This PR checks should failed

namse:
Now it should be passed

namse:
I removed `clean-up` because we don't use it now.

bigfoodK (approved):

PR #281, opened 2022-07-20
Use children shader like image

namse:
From now on, namui shader can evaluate other shader in shader.
https://skia.org/docs/user/sksl/#evaluating-sampling-other-skshaders

I put the sample which scroll image using this feature.

Image Sample from https://www.youtube.com/watch?v=jBHSDxNVoJg&ab_channel=%ED%9A%A8%ED%9A%A8ONTV. Thank you 효효ONTV!


namui/src/namui/skia/base/paint_builder.rs
```diff
@@ -19,10 +18,22 @@ pub struct PaintBuilder {
...
+    paint_style: Option<PaintStyle>,
+    anti_alias: Option<bool>,
+    stroke_width: Option<Px>,
+    stroke_cap: Option<StrokeCap>,
+    stroke_join: Option<StrokeJoin>,
+    color_filter: Option<(Color, BlendMode)>,
+}
+
+static PAINT_CACHE: OnceCell<Mutex<lru::LruCache<PaintBuilderWithoutShader, Arc<Paint>>>> =
+    OnceCell::new();
```
namse, line 36: I changed paint cache strategy.

paint won't cache using shader. paint just set shader everytime when they pop from cache. If the shader is the same with previous, they don't set shader.

namse: This is because shader could be changed too often against other paint properties. for example, when I make scroll, it create shader with time in every frame.

Later I will prevent creating shader in render stage, but only in draw stage. they don't have to create shader in rendering stage.

namse: Issue tracked in https://github.com/NamseEnt/namseent/issues/282

bigfoodK (approved):

PR #283, opened 2022-07-20
Add vector types system like xy and wh

namse:
It become powerful and useful, especially for shader programming.


namse:
@bigfoodK If you think it should have a test code, let me know then I will make some test codes.

bigfoodK (approved):
If there was an existing test, the test is needed.
But it doesn't seem to have existed before.
If there is a problem later, I think we can add a test code for the problem.

PR #294, opened 2022-08-01
Support deep link

bigfoodK:
# Before merging it

- [x] #292 needs to be merged fisrt

# How to use deeplink
Create a `.namuideeplink` file in the project root. The project root is where `.namuibundle` is located.

## `.namuideeplink` file
- One scheme per line.
- You can write comments with #.
ex)
```
luda-editor
# this line and
# this line will ignored.
other-scheme
```
will receive `luda-editor://some.url` and `other-scheme://other.url`

### DIFF
--- a/namui-cli/electron/src/main.js
+++ b/namui-cli/electron/src/main.js
@@ -6,8 +6,6 @@ const path = require("path");
 const config = getConfig();
 const gotTheLock = app.requestSingleInstanceLock();
 
-let mainWindow;
-
 if (!gotTheLock) {
     app.quit();
 } else {
@@ -28,7 +26,7 @@ if (!gotTheLock) {
 }
 
 function createWindow() {
-    mainWindow = new BrowserWindow({
+    const mainWindow = new BrowserWindow({
         width: 1280,
         height: 720,
         webPreferences: {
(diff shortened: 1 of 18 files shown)


namui-cli/electron/src/main.js
```diff
@@ -1,17 +1,41 @@
...
+    });
+
+    app.on("window-all-closed", function () {
+        if (process.platform !== "darwin") app.quit();
+    });
+}
 
 function createWindow() {
-    const mainWindow = new BrowserWindow({
+    mainWindow = new BrowserWindow({
```
namse, line 31: why const removed?

namse (changes requested):

namui-cli/electron/src/main.js
bigfoodK: It's weird...
It doesn't seem to have to be a global variable. I think I tried it during prototyping. I will revert it.
https://github.com/NamseEnt/namseent/pull/294/files/6bd08fa77e244bc5017495252b8a3d0b33d143a1#diff-bd152e97aedadf849d265dabab4438788196f4db0395d988755a138c5476ee9cR9

namse (approved):

PR #295, opened 2022-08-01
Support open external feature

bigfoodK:
# Before merging it

- [x] #292 needs to be merged first
- [x] #294 needs to be merged first

# Behavior
## on electron
Same as `shell.openExternal(url)`.
In most cases, it opens with a default browser rather than an electron.

## on non-electron
Same as ` web_sys::window().unwrap().open_with_url_and_target(url, "_blank");`

### DIFF
--- a/namui-cli/electron/src/main.js
+++ b/namui-cli/electron/src/main.js
@@ -1,4 +1,4 @@
-const { app, BrowserWindow, ipcMain } = require("electron");
+const { app, BrowserWindow, ipcMain, shell } = require("electron");
 const isDev = require("electron-is-dev");
 const { existsSync, readFileSync } = require("fs");
 const path = require("path");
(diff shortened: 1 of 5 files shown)


namui-cli/electron/src/main.js
```diff
@@ -1,17 +1,42 @@
-const { app, BrowserWindow, ipcMain } = require("electron");
+const { app, BrowserWindow, ipcMain, shell } = require("electron");
 const isDev = require("electron-is-dev");
+const { existsSync, readFileSync } = require("fs");
 const path = require("path");
 
 const config = getConfig();
+const gotTheLock = app.requestSingleInstanceLock();
+
+let mainWindow;
```
bigfoodK, line 9: [Support deep link](https://github.com/NamseEnt/namseent/pull/295/commits/39ec3672249ea417cd309bea6129d73a43862fa5)
commit on this branch is outdated. I will fix it

PR #296, opened 2022-08-01
Disable web security on electron

bigfoodK:
# Before merging it

- [x] #292 needs to be merged first
- [x] #294 needs to be merged first

# Why disabled it?
Sometimes we need to bypass the cors. This is the simplest of many ways.
If there is a problem, we will have to use another method.

### DIFF
--- a/namui-cli/electron/src/main.js
+++ b/namui-cli/electron/src/main.js
@@ -32,6 +32,7 @@ function createWindow() {
         height: 720,
         webPreferences: {
             preload: path.join(__dirname, "preload.js"),
+            webSecurity: false,
         },
     });
     setOpenUrlEventHandler(app, mainWindow);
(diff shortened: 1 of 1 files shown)


namui-cli/electron/src/main.js
```diff
@@ -33,6 +33,7 @@ function createWindow() {
         height: 720,
         webPreferences: {
             preload: path.join(__dirname, "preload.js"),
+            webSecurity: false,
```
namse, line 36: please set `allowRunningInsecureContent` false to prevent other security problem.
image

> https://www.electronjs.org/docs/latest/api/browser-window#new-browserwindowoptions

bigfoodK: Ah! got it

bigfoodK: [b0d124f](https://github.com/NamseEnt/namseent/pull/296/commits/b0d124f4e989c3f8a1f7c3833f3b64e152e80dc8) fixed

namse (approved):

PR #286, opened 2022-07-22
Use github api for luda editor

bigfoodK:
# What I did
## Use github api instead of luda-editor-server
Read and write sequences and images with github api.
In this pr, user must enter the github authentication token to text input manually.
Authentication ui will be implemented in next pr.

I used my private repository as storage for develop.
Before merge this pr, It must be replaced by a repository of namseent.

## Remove luda-editor-server and luda-edior-nrpc
You need to run `git clean -f .` after merge this pr.

## TODO

- [ ] Decide which repository to use as a storage
- [ ] Decide which oauth client to use for oauth login

PR #297, opened 2022-08-02
Login to github with oauth

bigfoodK:
Use github oauth web flow.
It only works with electron.
I have tested only in the wsl-windows-electron environment.


luda-editor/luda-editor-client/src/app/authentication/authentication.rs
```diff
@@ -4,49 +4,57 @@ use namui_prebuilt::simple_rect;
 use std::sync::Arc;
 use wasm_bindgen_futures::spawn_local;
 
+const CLIENT_ID: &str = "Iv1.7b68d8d6f7b401df";
+const CLIENT_SECRET: &str = "350c026f799464cc1f0cb080325ba666d758e06d";
+
```
bigfoodK, line 9: Client_secret should secret... but...

luda-editor/luda-editor-client/src/app/app.rs
```diff
@@ -99,7 +103,8 @@ enum AppStage {
     },
 }
 
-fn initialize_app(github_api_client: Arc<GithubApiClient>) {
+fn initialize_app(_user_id: u32, github_api_client: Arc<GithubApiClient>) {
+    // TODO: use user_id
```
bigfoodK, line 107: Someone will use this user_id.

bigfoodK:
Uh...

namse:
Why it merged?

namse:
Ah I found the reason. Only master branch is protected with code reivew.

image

bigfoodK:
It is not mandatory because we have already discussed enough.
However, the oauth client should be replaced. I added this to TODO of #286.

PR #298, opened 2022-08-07
Simplify namui project's start fn

namse:
# What I did 
- Remove wasm-bindgen from namui project.
  - It makes namui project more abstracted.
- Add runtime project which make runtime environment like wasm-bindgen.
- Fix namui-cli using dist option
- Tie duplicated code in namui-cli with WasmWatchBuildService

# What exactly changed - summary

## Before

namui project exposes their fn start using wasm-bindgen macro.

## After

namui project only exposes just `pub async fn main`. namui-cli will make `runtime project`, it will uses user's namui project and call `{user-namui-project}::start.await`. Runtime project will exposes start function to browser, or just run in rust binary in native platform.

zulip: https://namseent.zulipchat.com/#narrow/stream/332914-.EA.B8.B0.EC.88.A0.EC.97.B0.EA.B5.AC.EC.86.8C/topic/runtime_project


namui-cli/src/services/rust_build_service.rs
```diff
@@ -124,12 +124,18 @@ impl CancelableBuilder {
...
+                                .join()
+                                .expect("fail to get stdout from thread");
+
+                            if cargo_outputs.is_empty() {
+                                return Err(format!(
+                                    "cargo build failed {stderr}",
+                                    stderr = stderr_reading_thread.join().unwrap()
+                                )
+                                .into());
+                            }
```
namse, line 137: I just found a new bug about this code. If it failed to run cargo, cargo output's length become 0, but previous code didn't check and just return Ok.

namui-cli/src/services/wasm_watch_build_service.rs
```diff
@@ -0,0 +1,120 @@
...
+use crate::{cli::Target, debug_println, util::print_build_result};
+use std::{error::Error, path::PathBuf, sync::Arc};
+
+pub struct WasmWatchBuildService {}
+
+pub struct WatchAndBuildArgs {
+    pub project_root_path: PathBuf,
+    pub port: u16,
+}
+impl WasmWatchBuildService {
```
namse, line 17: I just thought all code is same among wasm_{windows/linux}_electron and wasm_unknown_web. Please confirm this.

namui/Cargo.toml
```diff
@@ -21,7 +18,6 @@ lazy_static = "1.4.0"
 once_cell = "1.8.0"
 serde_repr = "0.1"
 tokio = { version = "1", default-features = false, features = ["sync"] }
-console_error_panic_hook = { version = "0.1.6", optional = true }
```
namse, line 24: Moved to runtime project.

namui-cli/src/services/wasm_watch_build_service.rs
bigfoodK: Good. I think it's okay.

bigfoodK (approved):

PR #299, opened 2022-08-07
Advance network http

namse:
# What I did

1. Make get, post, delete, put, patch method in network::http
2. Wrap `reqwest::Response` into `namui::network::http::Response` to provide apis in our flavor.


namui/src/namui/utils/simple_error_impl.rs
```diff
@@ -0,0 +1,11 @@
+#[macro_export]
+macro_rules! simple_error_impl {
```
namse, line 2: This is additional macro which provide making error impl for struct very easily.

bigfoodK (approved):

PR #300, opened 2022-08-07
Get project name from manifest

namse:
Previously I tried to get project name using directory name. It's wrong. Real project name is written in Cargo.toml. So I read it and used it.


namui-cli/src/services/runtime_project/wasm.rs
```diff
@@ -1,9 +1,11 @@
+use std::path::PathBuf;
+
```
bigfoodK, line 2: Here is empty line

bigfoodK (changes requested):

namui-cli/src/services/runtime_project/wasm.rs
namse: I found that rust analyzer automatically make a new line.
https://rust-analyzer.github.io/manual.html#auto-import

I will try coding without this option in other branch and if it works well then I will make PR later about option.

bigfoodK (approved):

PR #302, opened 2022-08-08
Start browser or electron after first build

namse:
Some code should be not run.
When I change the code and run `namui start`, it will run previously built code which may not be able to run.

So I changed namui-cli to run browser or electron after first build.


namse:
Here is new idea, what about start `WasmBundleWebServer` after first build?

If we just make `WasmBundleWebServer` but don't run it, then after first build if we start `WasmBundleWebServer`, it would be better than current PR's solution.

bigfoodK:
Why do you think new idea is better @namse ?

namse:
Because there is one edge case for current pr. if user already opened browser, bundle server will serve previous build to them.

bigfoodK:
Then I agree with the second idea.

namse:
Let me do second solution later. I think it can be done next time. let's merge this pr first.

bigfoodK (approved):

PR #206, opened 2022-06-07
Prevent arrow & space key for scroll & history move

namse:
I found that Alt+ArrowLeft activate `go back` button. I prevent it with code clean (https://github.com/NamseEnt/namseent/issues/185)

PR #307, opened 2022-08-10
Only prevent default on alt key

namse:
Related PR: #206 
Related Zulip: https://namseent.zulipchat.com/#narrow/stream/332914-.EA.B8.B0.EC.88.A0.EC.97.B0.EA.B5.AC.EC.86.8C/topic/Fix.20text-input.20space

I found that key-down's preventing default block input space and arrow keys. I fixed it.

Check list
- It still block alt+arrow key
- It doesn't block space and arrow key in text input

### DIFF
--- a/namui/src/namui/system/keyboard.rs
+++ b/namui/src/namui/system/keyboard.rs
@@ -45,17 +45,8 @@ impl KeyboardSystem {
                         let mut pressing_code_set = pressing_code_set.write().unwrap();
                         pressing_code_set.insert(code);
 
-                        match code {
-                            Code::Space
-                            | Code::ArrowLeft
-                            | Code::ArrowRight
-                            | Code::ArrowUp
-                            | Code::ArrowDown
-                            | Code::AltLeft
-                            | Code::AltRight => {
-                                event.prevent_default();
-                            }
-                            _ => {}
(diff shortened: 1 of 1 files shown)


namui/src/namui/system/keyboard.rs
```diff
@@ -45,17 +45,8 @@ impl KeyboardSystem {
                         let mut pressing_code_set = pressing_code_set.write().unwrap();
                         pressing_code_set.insert(code);
 
-                        match code {
-                            Code::Space
```
bigfoodK, line 49: Does scrolling occur when the space bar is pressed?

namse: I'm not sure why I did care about scrolling. We don't have scroll-able element, isn't it?

bigfoodK: Yes, we only use the canvas that fits exactly the size of the viewport

bigfoodK: Then I think it's ok to merge

namse: I think the main purpose of #206 is alt+left or alt+right.

bigfoodK (approved):

PR #328, opened 2022-09-11
Apply on_top filter to text_input

namse:
# What Happened?
- `on_top`'s mouse event filtering doesn't worked with text_input
  - Because `on_top` mouse filtering was on [`call_mouse_event`](https://github.com/NamseEnt/namseent/blob/master/namui/src/namui/render/rendering_tree/mod.rs#L214), which called after [`text_input`'s `on_mouse_down_in`](https://github.com/NamseEnt/namseent/blob/03997970ec349e6babb1e44d02ca9d176e63805f/namui/src/namui/namui_context/main_loop/post_update_and_render.rs#L27).
- So I opened modal and clicked it, the text input behind the modal was clicked even I stop propagation.

# How did I fix
- Called `text_input`'s mouse down event handler with `text_input` rendering tree's `attach_event`. So it finally works with `RenderingTree`'s `call_mouse_event` logic.

Tested in visual novel editor and namui text_input sample

### DIFF
--- a/namui/src/namui/system/text_input/mouse_event.rs
+++ b/namui/src/namui/system/text_input/mouse_event.rs
@@ -1,51 +1,50 @@
 use super::*;
 
-pub(crate) fn on_mouse_down_in(namui_context: &NamuiContext, raw_mouse_event: &RawMouseEvent) {
+pub(crate) fn on_mouse_down_in_before_attach_event_calls() {
+    *TEXT_INPUT_SYSTEM.dragging_text_input_id.lock().unwrap() = None;
+}
+
+pub(crate) fn on_mouse_down_in_after_attach_event_calls() {
+    if TEXT_INPUT_SYSTEM
+        .dragging_text_input_id
+        .lock()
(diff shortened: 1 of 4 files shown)


namui/src/namui/system/text_input/mouse_event.rs
```diff
@@ -1,51 +1,50 @@
 use super::*;
 
-pub(crate) fn on_mouse_down_in(namui_context: &NamuiContext, raw_mouse_event: &RawMouseEvent) {
```
namse, line 3: I divide this function into 3 parts.
- Which call before text_input's `attach_event`'s calls
- Which will called at text_input's `attach_event`
- Which call after text_input's `attach_event` calls

So I assumed that `TEXT_INPUT_SYSTE.dragging_text_input_id` will be `None` if any text_input's `attach_event`s wasn't called. At that time I reset `last_focused_text_input_id` together.

bigfoodK (approved):

PR #330, opened 2022-09-15
Support Keydown, Focus for text_input

namse:
# What I did

- Support keydown event like react
  - Usage Example) you can stop input when `Enter` key down.
- Focus text input by api
  - Usage Example) You can focus text input when you open modal.
  - Current known issue: focus function may keep wrong selection.
- `enum Event` become nested.

### DIFF
--- a/luda-editor/client/src/pages/sequence_edit_page/loaded/update/mod.rs
+++ b/luda-editor/client/src/pages/sequence_edit_page/loaded/update/mod.rs
@@ -31,26 +31,21 @@ impl LoadedSequenceEditorPage {
                 }
             }
         } else if let Some(event) = event.downcast_ref::<text_input::Event>() {
-            match event {
-                text_input::Event::Focus(_)
-                | text_input::Event::Blur(_)
-                | text_input::Event::SelectionUpdated(_) => {}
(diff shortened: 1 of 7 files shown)


luda-editor/client/src/pages/sequence_edit_page/loaded/update/mod.rs
```diff
@@ -31,26 +31,21 @@ impl LoadedSequenceEditorPage {
                 }
             }
         } else if let Some(event) = event.downcast_ref::<text_input::Event>() {
-            match event {
-                text_input::Event::Focus(_)
-                | text_input::Event::Blur(_)
-                | text_input::Event::SelectionUpdated(_) => {}
```
bigfoodK, line 37: Is it deprecated to use all event enum variations in the match statement to prepare for addition of event enum variation?

bigfoodK (approved):

luda-editor/client/src/pages/sequence_edit_page/loaded/update/mod.rs
namse: Not deprecated, just made sure that this code only use this event. I don't have a confident opinion about this.

PR #334, opened 2022-09-20
Support i32 for table, title for typography

namse:
# What I did

1. Support i32 for table ratio
2. title for typography
  - 20px bold
3. `typography::left` only needs height, not wh.


namui-prebuilt/src/typography/mod.rs
```diff
@@ -0,0 +1,47 @@
...
+            ..Default::default()
+        },
+        max_width: None,
+    })
+}
+
+fn adjust_font_size(height: Px) -> IntPx {
+    // 0, 4, 8, 16, 20, ...
+    let mut font_size: Px = height * 0.7;
+    font_size -= font_size % 4;
```
bigfoodK, line 30: It's beautiful.

bigfoodK (approved):

PR #342, opened 2022-09-23
Allow dead_code for canvas_kit

namse:
We had a error on github action, deploying luda editor to aws.

image

It wasn't reproduced in my local computer, so I couldn't figured out it.

I put allow for dead_code in canvas_kit module, it's ok to put that. at that module I only put canvas_kit external library things.

Deployment has been successful after allow dead_code.

image


bigfoodK:
On my system, Ii can be reproduced. I think it's time to update.
```
rustup 1.25.1
rustc 1.64.0
```

bigfoodK (approved):

namse:
AH MY GOD rust new version!?

PR #350, opened 2022-09-27
Fix rust version 1.64

namse:
I had a problem that `namui test` docker uses rust version 1.58.
image

let's fix the rust version of docker.


namse:
I think `docker pull` will fix this problem. So I close this for now.

namse:
Hmm. no, docker pull doesn't help the problem.

bigfoodK (approved):

PR #355, opened 2022-09-30
Preview visual novel sequence

namse:
# What I did

- visual novel sequence player

You can find visual novel player on top-right of sequence editor.

You can check in here https://wvtr66c5qnuen5tzndzwbdpcj40psacc.lambda-url.ap-northeast-2.on.aws/


luda-editor/client/src/components/sequence_player.rs
```diff
@@ -0,0 +1,312 @@
...
+                let transition_duration = 500.ms();
+
+                let delta_time = Time::now() - *start_time;
+                if delta_time > transition_duration {
+                    self.state = State::ShowingCut {
+                        cut_index: *from_cut_index + 1,
+                    };
+                } else {
+                    *transition_progress = (delta_time / transition_duration).one_zero();
+                }
```
bigfoodK, line 101: It looks good to me, but is there a reason for calculating `transition_progress` in update not render?
`transition_progress` is not updated by specific event, but rather indicates how it will be rendered over time.

bigfoodK (approved):

luda-editor/client/src/components/sequence_player.rs
namse: To make sure transition_progess is always less than 1 on render. If other update job takes too much time, it would be bigger than 1 at render time.
Yeh, it's very rare case but I just thought about it.

PR #356, opened 2022-09-30
Use docker-cli, not docker compose v2

namse:
I hate docker doesn't install compose v2 automatically


bigfoodK (approved):
Oh, I didn't know that
image

PR #359, opened 2022-10-02
Impl prototype of luda adventure

bigfoodK:
# Features
- Game system.
  - Use entity and component concepts.
- Camera.
  - In this version, camera will always follow character.
- Guide icon.
- Character.
- Map.
  - Use tile concept.

# Before merge

- [x] https://github.com/NamseEnt/namseent/pull/350
- [x] Rmove auto_ops from cargo.toml https://github.com/NamseEnt/namseent/pull/378
- [x] Merge #379


luda-adventure/src/app/game/types/game_object/types/positioner/heading.rs
```diff
@@ -0,0 +1,24 @@
...
+
+pub fn get_heading_from_velocity(velocity: Velocity) -> Option<Heading> {
+    let unit_delta_x = velocity.x * 1.ms();
+    let unit_delta_y = velocity.y * 1.ms();
+
+    if unit_delta_x == 0.tile() {
+        return None;
+    }
+    let tangent = unit_delta_y / unit_delta_x;
+    if tangent.abs() > 8.0 {
```
bigfoodK, line 17: image
*abs(tan) > 8

namui/src/namui/common/types/macros.rs
```diff
@@ -76,7 +80,20 @@ macro_rules! common_for_f32_type {
         impl<T: $crate::types::Ratio> std::ops::Mul<T> for $your_type {
             type Output = $your_type;
             fn mul(self, rhs: T) -> Self::Output {
-                (self.0 * rhs.as_f32()).into()
+                (self.as_f32() * rhs.as_f32()).into()
```
bigfoodK, line 83: @namse We need discussion.
`Percent` act like [this](https://github.com/NamseEnt/namseent/blob/38e13c985a7d8771e9f280beb6edbe5a10674b2d/namui/src/namui/common/types/percent.rs#L4-L9). 
```rust
common_for_f32_type!(
    Percent,
    // To f32
    |lhs: Percent| -> f32 { lhs.0 / 100.0 },
    // From f32
    |lhs: f32| -> Percent { Percent(lhs * 100.0) },
    percent,
    PercentExt
);
```
It leads to situation like below.
`100 percent * 4 = 40,000 percent`, because actual behavior is
1. `(100 * 4).into()`
2. `400.into()`
3. `Percent(400 * 100)`
4. `Percent(40,000)`

Which one do you think is better?
1. Use `as_f32` while impl `Mul` (current)
5. Store without multiplying by 100 at `Percent.0`

namse: I'm not sure what are you talking about. Is this test cover what you concern?
image

bigfoodK: Yes, that test shows 60,000, not 600.

namse: What do you mean? this test shows `600 percent` is `150 percent * 4.0`

bigfoodK: Test was failed before a9da757266e1ad95539ebe4e2429a854640582cd
There was problem. Percent stores 100 multiplied f32 at `Percent.0`.
But previous code used `Percent.0` directly for Mul. So test was failed with 60,000, not 600.
image

namse: All test include new test are passed well in master branch. I have no idea what you are saying.
image

namse: Give me a test code and code branch which fail the test

namse: I think it's because of `impl Ratio for f32`
https://github.com/NamseEnt/namseent/blob/a9da757266e1ad95539ebe4e2429a854640582cd/namui/src/namui/common/types/ratio.rs#L10

namse: Ok, `assert_eq!(600.percent(), 150.percent() * 400.percent());` is failed in master branch.

It should be change to f32 and change back again. It's not a good idea to store Percent without multiplying 100, because it should not the limit how user want to save data inside of struct.

Simply put as_f32 there please.
```rust
impl<T: $crate::types::Ratio> std::ops::Mul<T> for $your_type {
    type Output = $your_type;
    fn mul(self, rhs: T) -> Self::Output {
        (self.as_f32() * rhs.as_f32()).into()
                // ^--this
    }
}
```

namse:
I divided namui changes to pr https://github.com/NamseEnt/namseent/pull/378

luda-adventure/src/app/game/types/game_object/types/positioner/heading.rs
namse: What's the purpose of this gray dead-zone?

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
```diff
@@ -0,0 +1,156 @@
...
+    }
+    pub fn get_predicted_movement_end_time(&self) -> Time {
+        self.movement_plan.get_predicted_movement_end_time()
+    }
+    pub fn predict_movement(
+        &mut self,
+        collider: &Collider,
+        collision_box_list: &[crate::app::game::CollisionBox],
+    ) {
+        let prediction_start_time = self.get_predicted_movement_end_time();
```
namse, line 37:
```rust
let prediction_start_time = self.get_predicted_movement_end_time();
                                                      // ^-- `end` is right?
```

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
```diff
@@ -0,0 +1,156 @@
...
+                    CollisionDirection::Vertical => Xy {
+                        x: self.movement_plan.directed_movement.velocity.x,
+                        y: Per::new(0.tile(), 1.ms()),
+                    },
+                    CollisionDirection::Horizontal => Xy {
+                        x: Per::new(0.tile(), 1.ms()),
+                        y: self.movement_plan.directed_movement.velocity.y,
+                    },
+                };
+                let second_collision = predict_collision(
```
namse, line 86: Why this function calculates two collision at once?

luda-adventure/src/app/game/types/game_object/types/positioner/predict_collision.rs
```diff
@@ -0,0 +1,168 @@
...
+            };
+            let (collision_start_remaining_time, collision_end_remaining_time) =
+                match collision_direction {
+                    CollisionDirection::Horizontal => (
+                        intersection_start_remaining_time_of_x,
+                        intersection_end_remaining_time_of_y,
+                    ),
+                    CollisionDirection::Vertical => (
+                        intersection_start_remaining_time_of_y,
+                        intersection_end_remaining_time_of_x,
```
namse, line 80: is it right that `collision_end_remaining_time` is end of x or y? not the `start` of counter-side?

luda-adventure/src/app/game/update/handle_character_movement_on_key_event.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub fn handle_character_movement_on_key_event(&mut self, current_time: Time) {
+        if let Some((_character, (_player_character, positioner))) = self
+            .ecs_app
+            .query_entities_mut::<(&PlayerCharacter, &mut Positioner)>()
+            .first_mut()
+        {
+            let character_velocity = get_character_velocity_from_key_state();
+            let character_velocity_has_not_changed = character_velocity.x * 1.ms()
+                == self.state.character.last_velocity.x * 1.ms()
+                && character_velocity.y * 1.ms() == self.state.character.last_velocity.y * 1.ms();
```
namse, line 14: Why do you multiply 1.ms() instead of just comparing Per struct?

luda-adventure/src/app/game/types/game_object/types/positioner/heading.rs
bigfoodK: There is no important reason. 
When I was playing Among Us with a joystick, it bothered me to see it flip from side to side often while moving up.

luda-adventure/export-known-ids-macro/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+        }
+    });
+
+    TokenStream::from(quote! {
+        use namui::prelude::uuid;
+
+        #(#exports)*
+    })
+}
+
```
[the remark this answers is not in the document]
bigfoodK: This macro checks duplicated uuid and emit error at compile time.
This prevents developers from making mistakes. So I think I need.

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
bigfoodK: Yes, next prediction should start from end of previous prediction.

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
bigfoodK: Because of float, it was ambiguous to determine whether it was in a collision.
So there was a case of getting infinitely close to the wall.
So I decided to predict post-collision movements instead of checking if they were in a collision.

luda-adventure/src/app/game/game.rs
```diff
@@ -0,0 +1,133 @@
+use super::{player_character::new_player, *};
+use namui::prelude::*;
+
+pub struct Game {
+    pub player_entity_id: Uuid,
```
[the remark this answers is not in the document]
bigfoodK: You mean, don't use `const SOME_ID = "12d12-..."`.

luda-adventure/src/app/game/types/game_state/character/character_state.rs
```diff
@@ -0,0 +1,14 @@
...
+    pub last_velocity: Velocity,
+}
+
+impl CharacterState {
+    pub fn new() -> Self {
+        Self {
+            last_velocity: Velocity::single(Per::new(0.tile(), 1.ms())),
+        }
+    }
+}
```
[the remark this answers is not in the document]
bigfoodK: That's good idea. I will fix it.

luda-adventure/src/app/game/update/handle_character_movement_on_key_event.rs
bigfoodK: Just because `Ord` is not implemented for `Per`.

luda-adventure/src/app/game/types/game_object/types/positioner/predict_collision.rs
```diff
@@ -0,0 +1,168 @@
...
+        speed.invert() * target_less_side_minus_character_greater_side;
+    let intersection_prediction = match (speed * 1.0.ms())
+        .partial_cmp(&0.0_f32.tile())
+        .expect("Invalid speed")
+    {
+        std::cmp::Ordering::Less => IntersectionPrediction::WillIntersect {
+            start_remaining_time: target_greater_side_minus_character_less_side_divided_by_speed,
+            end_remaining_time: target_less_side_minus_character_greater_side_divided_by_speed,
+            start_remaining_distance: target_greater_side_minus_character_less_side,
+        },
```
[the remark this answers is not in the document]
bigfoodK: `speed * 1.0ms < 0.tile` means moving direction is neg, `speed * 1.0ms > 0.tile` means pos.
This is more about direction than intersect. I will clearfy it.

luda-adventure/src/app/game/types/game_object/types/positioner/predict_collision.rs
bigfoodK: Here is a design sketch for this.

image

luda-adventure/export-known-ids-macro/src/lib.rs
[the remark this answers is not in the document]
namse: I think you can just make a new const uuid in your code and use it. can't it?

luda-adventure/src/app/game/game.rs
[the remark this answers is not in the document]
namse: No, just like this

```rust
pub struct Game {
  ...
  pub quest_id: Uuid,
}

implk Game {
  pub fn new() -> Self {
    ...
    let quest_id = namui::uuid();
    
    Self {
      quest_id,
    }
  }
}
```

And then, it would be better to save quest id in json file, not code. Are you planing making all resources in code?

luda-adventure/src/app/game/update/handle_character_movement_on_key_event.rs
namse: Then what about impl that?

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
namse: What you mean is that it would be ambiguous to determine whether it was in a collision or not because of floating number error?

Do `second_collision` mean the second collision after the first collision so it moved by one direction then collide? like a, character moves up-right and collide on top and move right and collide on right-situation?

Could it be good solution for that cases...? I'm not sure about it. I just thought saving collision status instead of calculating every time whether it is in a collision pr not - because of floating number error.

luda-adventure/src/app/game/types/game_state/character/character_state.rs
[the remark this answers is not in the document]
bigfoodK: Fixed [44d77a9](https://github.com/NamseEnt/namseent/pull/359/commits/44d77a943a2abfe3f318d070656c4768d20aa224)

luda-adventure/src/app/game/update/handle_character_movement_on_key_event.rs
bigfoodK: Good. I will try that

luda-adventure/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
```
namse, line 1: Please add Ci test for adventure

luda-adventure/export-known-ids-macro/src/lib.rs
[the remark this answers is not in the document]
bigfoodK: Because you don't think it's going to happen that the uuid duplicated?

luda-adventure/src/app/game/update/handle_character_movement_on_key_event.rs
bigfoodK: [44d77a9](https://github.com/NamseEnt/namseent/pull/359/commits/44d77a943a2abfe3f318d070656c4768d20aa224)

luda-adventure/export-known-ids-macro/src/lib.rs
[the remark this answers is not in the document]
namse: Yes and I'm not sure saving data in code.

luda-adventure/src/app/game/types/game_object/types/positioner/predict_collision.rs
```diff
@@ -121,22 +121,43 @@ fn predict_intersection_on_one_axis(
...
+        Negative,
+    }
+    let movement_direction = match (speed * 1.0.ms())
+        .partial_cmp(&0.0_f32.tile())
+        .expect("Invalid speed")
+    {
+        std::cmp::Ordering::Less => MovementDirection::Negative,
+        std::cmp::Ordering::Equal => MovementDirection::Zero,
+        std::cmp::Ordering::Greater => MovementDirection::Positive,
+    };
```
namse, line 136: I hope `Per` implement [is_sign_negative](https://doc.rust-lang.org/std/primitive.f64.html#method.is_sign_negative) for this, because it looks not cool multiplying 1.0.ms() for calcultation.

luda-adventure/export-known-ids-macro/src/lib.rs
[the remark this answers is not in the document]
namse: The only way to make a couple of duplicated uuids is when you write uuid by your brain and hands. If you use uuid generator, such any website googled, you always make a non-duplicated uuid. So that mean you don't need to spend your time for manage this macro.

luda-adventure/src/app/game/game.rs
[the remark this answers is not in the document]
bigfoodK: Why do you think it's better?

Actually I will save resources to your code. I don't think there's any reason to save it in the json file because I didn't plan to make an editor.

luda-adventure/export-known-ids-macro/src/lib.rs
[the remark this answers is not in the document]
bigfoodK: `manage`... got it. let's remove this.

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
bigfoodK: 1. Yes
2. Yes
3. Hmmmmmm. I'm not sure this will reduce the complexity yet. Let me think for one more night.

luda-adventure/src/app/game/game.rs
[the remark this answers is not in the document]
namse: Then if somebody wants to make a quest then they need to read and write code?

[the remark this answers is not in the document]
namse: I mean, above what i'm saying is saving data in external file,

And What I want to say is that's hard-coding which generally one of the code-smell. There are very much of reason why it's not a good idea to make a hard-coding.

What I suggest is that making it flexible anytime that Id can be loaded from file or just generated from code like my above example code.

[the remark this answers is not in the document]
namse: Who will make a quest?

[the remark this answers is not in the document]
bigfoodK: I think I will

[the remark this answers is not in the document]
namse: I see. Anyway it's not a that much big deal but I think const id isn't good option for flexibility of code and putting it to state is better. Go a head what do you want to do.

luda-adventure/export-known-ids-macro/src/lib.rs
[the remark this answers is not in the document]
bigfoodK: [db487dc](https://github.com/NamseEnt/namseent/pull/359/commits/db487dc01071d200c454b7ee00140a9ee7e493cf)

luda-adventure/.cargo/config.toml
bigfoodK: [e100e0c](https://github.com/NamseEnt/namseent/pull/359/commits/e100e0cc5e14acffcf50803dcc0f047c3fdebfd4)

luda-adventure/src/app/game/types/game_object/types/positioner/predict_collision.rs
bigfoodK: #379

luda-adventure/src/app/game/update/handle_character_movement_on_key_event.rs
bigfoodK: #379

luda-adventure/src/app/game/types/game_object/types/positioner/predict_collision.rs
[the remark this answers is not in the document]
bigfoodK: [44d77a9](https://github.com/NamseEnt/namseent/pull/359/commits/44d77a943a2abfe3f318d070656c4768d20aa224)

luda-adventure/src/app/game/types/game_object/types/positioner/predict_collision.rs
```diff
@@ -121,22 +121,43 @@ fn predict_intersection_on_one_axis(
...
     let target_greater_side_minus_character_less_side = target_greater_side - character_less_side;
     let target_greater_side_minus_character_less_side_divided_by_speed =
         speed.invert() * target_greater_side_minus_character_less_side;
     let target_less_side_minus_character_greater_side = target_less_side - character_greater_side;
     let target_less_side_minus_character_greater_side_divided_by_speed =
         speed.invert() * target_less_side_minus_character_greater_side;
```
namse, line 142: I don't feel good with this code because these variable names doesn't have any meaning but just the right side's calculation.

Let's think about `let one_plus_two = 1 + 2;`. Why it need to make a new variable? Any good reason for it?

I think variable names are meaningful when it has different meaning against right-side calculation.

For example
```rust
let target_greater_side_minus_character_less_side = target_greater_side - character_less_side;
// would be better with
let distance_between_character_less_to_target_greater = target_greater_side - character_less_side;
```

namse (changes requested):

luda-adventure/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+
+[dependencies]
+export-known-ids-macro = {path = "./export-known-ids-macro"}
+ecs-macro = {path = "./ecs-macro"}
+namui = {path = "../namui"}
+namui-prebuilt = {path = "../namui-prebuilt"}
+once_cell = "1.15.0"
+rustc-hash = "1.1.0"
+serde = "1.0"
+auto_ops = "0.3"
```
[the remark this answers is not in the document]
namse: Please update this too @bigfoodK

luda-adventure/src/app/game/types/game_object/types/positioner/predict_collision.rs
bigfoodK: Ahh, I agree

bigfoodK: [80befd7](https://github.com/NamseEnt/namseent/pull/359/commits/80befd7b6f487deec5db35fec423c75dbc814ec4)

luda-adventure/Cargo.toml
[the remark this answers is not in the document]
bigfoodK: [3f9c869](https://github.com/NamseEnt/namseent/pull/359/commits/3f9c8692eb9bad38ebbb8503ea57fd4d0472239f)

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
bigfoodK: I don't think it's going to be much. Can you explain in more detail the logic you are thinking of?

namse: [As I said in zulip before](https://namseent.zulipchat.com/#narrow/stream/332914-.EA.B8.B0.EC.88.A0.EC.97.B0.EA.B5.AC.EC.86.8C/topic/.28no.20topic.29/near/298384104), If I make it, I will make a queue of movement status with enum.

```rust
enum Movement {
  FreeMove {
    from: Xy,
    to: Xy
  },
  CollideAndSlideToWall {
    start: Xy,
    wall_axis: XOrY,
  },
  CollideAndStuck {
    position: Xy,
  }
}

let movement_queue: Vec<Movement> = calculate_movement(from_time, to_time);

let current_position = calculate_position(time, &movement_queue);

fn calculate_movement(from_time: Time, to_time: Time) -> Vec<Movement> {
  ...
  if let Movement(CollideAndSlideToWall { start, wall_axis }) = last_movement {
    if there_is_new_wall_so_will_be_stuck_by_them {
      queue.push(Movement::CollideAndStuck { position })
    }
  }
}
```

namse: What you have a trouble is that your data structures don't have enough data to calculate or `express` what you want to do.

bigfoodK: I think I understand the word "saving collision status".
Logic has not changed much, but the number of collision calculations has decreased.
Readability is... I don't know.

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
```diff
@@ -119,8 +119,11 @@ impl Positioner {
                         .push(movement_until_collision);
                 } else {
                     let delta_time = last_collision.end_time - last_collision.start_time;
-                    let end_position =
-                        velocity_after_last_collision * delta_time + prediction_start_position;
+                    let end_position = prediction_start_position
+                        + Xy {
+                            x: velocity_after_last_collision.x * delta_time,
+                            y: velocity_after_last_collision.y * delta_time,
+                        };
```
namse, line 126: Why did you revert this?

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
```diff
@@ -32,14 +32,15 @@ impl Positioner {
...
-        collision_box_list: &[crate::app::game::CollisionBox],
+        target_collision_box_list: &[crate::app::game::CollisionBox],
     ) {
-        let prediction_start_time = self.get_predicted_movement_end_time();
-        let prediction_start_position = self.get_position(prediction_start_time);
+        let last_prediction = self.get_last_prediction();
+        let prediction_start_time = last_prediction.end_time;
+        let prediction_start_position = last_prediction.end_position;
         let character_collision_box = collider.get_collision_box(prediction_start_position);
+        let directed_movement = &self.movement_plan.directed_movement;
```
namse, line 41: What's the meaning `directed_movement` for ?

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
```diff
@@ -50,107 +51,116 @@ impl Positioner {
...
+        }
     }
 
     pub fn get_position(&self, current_time: Time) -> Xy<crate::app::game::Tile> {
         self.movement_plan
             .get_position(current_time)
             .expect("get_position() of PlayerCharacter called before prediction")
     }
+
+    fn get_last_prediction(&self) -> Movement {
```
namse, line 156: FYI: You can return reference or impl Copy for Movement

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
```diff
@@ -50,107 +51,116 @@ impl Positioner {
...
+        match self.movement_plan.predicted_movement_list.last() {
+            Some(movement) => movement.clone(),
+            None => Movement {
+                start_time: (f32::NEG_INFINITY).ms(),
```
namse, line 160: Why start_time here is -inf?

luda-adventure/src/app/game/types/game_object/types/positioner/movement.rs
```diff
@@ -53,10 +55,20 @@ impl Movement {
             start_position: position,
             end_position: position,
             velocity: Xy::single(Per::new(0.tile(), 1.ms())),
+            movement_state: MovementState::Stuck,
```
namse, line 58: I think stay-forever is not stucking. And I believe you have wrong naming for `movement_state`.
`movement_state` should be `collision_state` because it is meaning about what kind of state this movement has for collision, to calculate [what you said](https://github.com/NamseEnt/namseent/pull/359#discussion_r988536420)

namse (changes requested):

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
bigfoodK: Because binary_operator between tuple_types colide.

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
bigfoodK: Ideal movement to be executed without any collision.
User directed movement.

namse: It makes some confusing because it looks like about direction (--> , <--)

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
bigfoodK: Kind of fallback. Would it be better to just emit an error?

luda-adventure/src/app/game/types/game_object/types/positioner/movement.rs
bigfoodK: Oh

bigfoodK: [f6fc994](https://github.com/NamseEnt/namseent/pull/359/commits/f6fc994268644b4ef995d238b4a2401e3b73506d)

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
bigfoodK: [6982da7](https://github.com/NamseEnt/namseent/pull/359/commits/6982da7181a4446bd46f8e1e029a8020f267c7cb)

luda-adventure/src/app/game/types/game_object/types/positioner/movement.rs
bigfoodK: [1bb6880](https://github.com/NamseEnt/namseent/pull/359/commits/1bb68800a7d9dc124477986a6969d78875fa4d4b)

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
namse: At least this code should let anyone knows that you want to put this for fallback.

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
namse: requested_movement?

luda-adventure/src/app/game/types/game_object/types/positioner/movement.rs
```diff
@@ -55,13 +55,13 @@ impl Movement {
...
             velocity: Xy::single(Per::new(0.tile(), 1.ms())),
-            movement_state: MovementState::Stuck,
+            collision_state: CollisionState::Stuck,
         }
     }
 }
 
 #[derive(Clone, Debug)]
-pub enum MovementState {
+pub enum CollisionState {
```
namse, line 64: So, it's value should be about collision, not movement.

Here is my suggestion.

- NotCollide
- WillCollide
- CollideOneAxis
- CollideOneAxisButWillBeTwo
- CollideTwoAxis

luda-adventure/src/app/game/types/game_object/types/positioner/movement_plan.rs
```diff
@@ -0,0 +1,59 @@
+use super::{CollisionState, Movement};
+use crate::app::game::Tile;
+use namui::prelude::*;
+
+pub struct MovementPlan {
+    pub original_movement: Movement,
+    pub predicted_movement_list: Vec<Movement>,
```
namse, line 7: I suggest making sure about this data.


There are two type of plan,
- The plan the user or somebody want to move or expected
- The real movement plan which will be happen in future absolutely

And we strongly assumed that no item will be moved, but only character will be moved.

Then there is only two types of movement result,
1. fixed position (when it is not character or player didn't move)
2. moving (only for character when player lets character move)

So I believe we can make it simpler like this.
Think about it and let me your opinion.

```rust
enum Motion {
    Fixed(Xy),
    Moving(Vec<MovePath>)
}

struct MovePath {
    start_xy: Xy,
    start_time: Time,
    end_xy: Xy,
    end_time: Time,
}

struct Positioner {
    motion: Motion
}

impl Positioner {
    pub fn xy(&self, time: Time) -> Xy {
        match self.motion {
            Motion::Fixed(xy) => xy,
            Motion::Moving(paths) => {
                paths.iter().find(|path| {
                    path.start_time <= time && time <= path.end_time
                }).map(|path| {
                    let ratio = (time - path.start_time) / (path.end_time - path.start_time);
                    Xy {
                        x: path.start_xy.x + (path.end_xy.x - path.start_xy.x) * ratio,
                        y: path.start_xy.y + (path.end_xy.y - path.start_xy.y) * ratio,
                    }
                }).unwrap_else(|| unreachable!())
            }
        }
    }
    pub fn update_motion(&mut self, motion: Motion) {
        self.motion = motion;
    }
}

impl App {
    fn update(&mut self, event: &dyn std::any::Any) {
        if is_time_to_update_character_motion { // when user change input axis or it needs to update prediction by time spending
            let (character, character_positioner) = self.ecs_app.find_character_and_its_positioner();
            let next_motion = character.calculate_next_motion(&self.ecs_app);
            character_positioner.update_motion(next_motion);
        }
    }
}

[Component]
struct Character {
    user_input_axis: Xy,
}

impl Character {
    pub fn calculate_next_motion(&self, ecs_app: &EcsApp) -> Motion {
        // get all objects
        // and calculate them at here all.
    }
}
```

namse (changes requested):

luda-adventure/src/app/game/types/game_object/types/positioner/movement_plan.rs
bigfoodK: Your suggestion looks more appropriate for ecs and our design.

bigfoodK: `calculate_next_motion` requires not only character components but also positioners and colliders. Therefore, I think it is appropriate for `calculate_next_motion` to be implemented outside rather than on character components. I think this is similar to ECS's system.


```rust
enum Motion {
    Fixed(Xy),
    Moving(Vec<MovePath>)
}
```
But I like this concept, and I think I need a diet in my current logic, so I'm trying to apply your suggestion.

namse: What's the purpose to put that function outside of character component? I already thought about it so I put `ecs_app: &EcsApp` as parameter to query all of them.

bigfoodK: For simplicity.
If we implement a function in a character component, We need to query the entity that owns this component and re-query the components that it has. Can we say this is a function of the character component?

I think it should be a system that uses character, positioner, colider components.
image

namse: We right now have no system.

It's not a big deal to query inside of character component's function.

It would be better to put function inside of character component because it's the only place that function will be used and gathering together by interest is good practice in my thought.

bigfoodK: Good. It's not a big deal.  I'll focus on `gathering together by interest` and continue.

namse: Hm.... I think it's ok to seperate all function from component and put them into one directory like a `systems`. Because they have not only single component's system but also combination's one. I'm not sure about it. go a head what you want.

bigfoodK: It acts like a `system`, but the name will be use `update`. Because it is more suitable for namui.

luda-adventure/src/app/game/types/game_object/types/positioner/movement.rs
bigfoodK: No longer use this.

luda-adventure/src/app/game/types/game_object/types/positioner/positioner.rs
bigfoodK: No longer use this.

luda-adventure/src/app/game/game.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    pub fn update(&mut self, event: &dyn std::any::Any) {
+        let current_time = now();
+        if let Some(event) = event.downcast_ref::<NamuiEvent>() {
+            match event {
+                NamuiEvent::KeyDown(_) => {
+                    self.handle_user_input_for_character_movement(current_time)
+                }
+                NamuiEvent::KeyUp(_) => self.handle_user_input_for_character_movement(current_time),
```
namse, line 43: Event can be queued so current system's state could be different to last event. So I think you should use event, not system.keyboard. how do you think about it?

luda-adventure/src/app/game/types/game_object/types/collider.rs
```diff
@@ -0,0 +1,123 @@
...
+    collision_box: CollisionBox,
+    simplified_collision_box_list: &mut Vec<CollisionBox>,
+) {
+    let collision_box_area = get_collision_box_area(collision_box);
+    for simplified_collision_box in simplified_collision_box_list.iter_mut() {
+        let merged_collision_box =
+            simplified_collision_box.get_minimum_rectangle_containing(collision_box);
+        let merged_area = get_collision_box_area(merged_collision_box);
+        let summed_area = collision_box_area + get_collision_box_area(*simplified_collision_box);
+        let can_merge = merged_area == summed_area;
```
namse, line 44: This mean you don't merge ㄱ shaped boxes, right?

luda-adventure/src/app/game/update/calculate_next_motion_of_character/calculate_next_movement_of_character.rs
```diff
@@ -0,0 +1,279 @@
...
+pub fn calculate_next_movement_of_character(ecs_app: &mut ecs::App, max_duration: Time) {
+    let target_collider_list = ecs_app.query_entities::<(&Collider, &Positioner)>();
+    let target_collision_box_list = simplify_collision_box_list(
+        target_collider_list
+            .iter()
+            .filter_map(|(target_entity, (target_collider, target_positioner))| {
+                if target_entity.id() == PLAYER_CHARACTER {
+                    None
+                } else {
+                    Some(target_collider.get_collision_box(target_positioner.xy(0.ms())))
```
namse, line 25: Why it use 0.ms(), not the current time to calculate? It's kind of TODO?

luda-adventure/src/app/game/game.rs
```diff
@@ -0,0 +1,128 @@
...
+                | NamuiEvent::MouseMove(_)
+                | NamuiEvent::ScreenResize(_)
+                | NamuiEvent::Wheel(_)
+                | NamuiEvent::DeepLinkOpened(_) => (),
+            }
+        }
+
+        while need_to_calculate_next_motion_of_character(&self.ecs_app, current_time, NEAR_FUTURE) {
+            calculate_next_movement_of_character(&mut self.ecs_app, NEAR_FUTURE);
+        }
```
namse, line 56: I think game don't have to check whether character needs to calculate next motion or not. it's not the responsibility for game.

This also makes performance poor because it query on every loop for each `need_to_calculate_next_motion_of_character` and `calculate_next_movement_of_character`, including collision merging.

What about letting them do this inside?

luda-adventure/src/app/game/update/calculate_next_motion_of_character/calculate_next_movement_of_character.rs
```diff
@@ -0,0 +1,279 @@
...
+            velocity,
+            character_collision_box,
+            &target_collision_box_list,
+        ) {
+            Some(remaining_collision_start_time) => {
+                Time::min(end_time, current_time + remaining_collision_start_time)
+            }
+            None => end_time,
+        };
+        let delta_time = end_time - current_time;
```
namse, line 60: I guess above end_time is not the real end_time until before you calculate delta_time here. that should be max_time or something like that.

luda-adventure/src/app/game/update/calculate_next_motion_of_character/calculate_next_movement_of_character.rs
```diff
@@ -0,0 +1,279 @@
...
+        let delta_xy = Xy {
+            x: velocity.x * delta_time,
+            y: velocity.y * delta_time,
+        };
```
namse, line 64: I made you velocity * delta_time works didn't I...?

luda-adventure/src/app/game/update/calculate_next_motion_of_character/calculate_next_movement_of_character.rs
```diff
@@ -0,0 +1,279 @@
...
+        let end_xy = character_xy + delta_xy;
+        let next_movement_path = MovementPath::new(current_time, end_time, character_xy, end_xy);
+
+        character_positioner.push_movement_path(next_movement_path);
```
namse, line 68: Does this system push movement to character positioner without clear? it will make out-of-memory problem and performance decrease for iteration.

luda-adventure/src/app/game/update/calculate_next_motion_of_character/calculate_next_movement_of_character.rs
```diff
@@ -0,0 +1,279 @@
...
+};
+use float_cmp::{ApproxEq, F32Margin};
+use namui::prelude::*;
+
+const COLLISION_MARGIN: F32Margin = F32Margin {
+    ulps: 1,
+    epsilon: 0.0001,
+};
+
+pub fn calculate_next_movement_of_character(ecs_app: &mut ecs::App, max_duration: Time) {
```
namse, line 16: I'm not sure yet what do you want to do with this. I know the logic but I feel it's very complicated.

Positioner, it seems having two state, Fixed and Dynamic, and most of code is for Dynamic case. I feel this very strange and i'm not clear with it.

luda-adventure/src/app/game/game.rs
bigfoodK: That makes sense. I will fix it.

luda-adventure/src/app/game/types/game_object/types/collider.rs
bigfoodK: Yes, actually this shape.
```
■
  ■
```
Outdated

luda-adventure/src/app/game/update/calculate_next_motion_of_character/calculate_next_movement_of_character.rs
bigfoodK: Yes, due to rust's borrow strategy.
Outdated

luda-adventure/src/app/game/update/calculate_next_motion_of_character/calculate_next_movement_of_character.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    if let Some((_character_entity, (character, character_collider, character_positioner))) =
+        ecs_app
+            .query_entities_mut::<(&PlayerCharacter, &Collider, &mut Positioner)>()
+            .first_mut()
+    {
+        let current_time = character_positioner.movement_end_time();
+        let character_xy = character_positioner.xy(current_time);
+        let mut velocity = character.velocity();
+        let mut end_time = current_time + max_duration;
```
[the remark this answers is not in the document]
bigfoodK: If it were `next_movement_start_time` and `next_movement_maximum_end_time`, it would be easier to read.
Outdated

luda-adventure/src/app/game/game.rs
bigfoodK: Isn't it easier to read? If not, who do you think should be responsible?

I don't think we need to seriously consider optimization now that we're making prototypes.

luda-adventure/src/app/game/update/calculate_next_motion_of_character/calculate_next_movement_of_character.rs
bigfoodK: I think so.

bigfoodK: Outdated

luda-adventure/src/app/game/update/calculate_next_motion_of_character/calculate_next_movement_of_character.rs
bigfoodK: It is flushed when user input is changed.
Outdated.

luda-adventure/src/app/game/update/calculate_next_motion_of_character/calculate_next_movement_of_character.rs
bigfoodK: It is another case. `Output` is not Lhs.
It is not like `Xy<Per<Tile, Time>>` * `Ratio` = `Xy<Per<Tile, Time>>`.
It should be like `Xy<Per<Tile, Time>>` * `Time` = `Xy<Tile>`.

luda-adventure/src/app/game/update/calculate_next_motion_of_character/calculate_next_movement_of_character.rs
bigfoodK: I think so too. But I couldn't find a better way.
Outdated

luda-adventure/src/app/game/update/resolve_collision_about_character.rs
```diff
@@ -0,0 +1,129 @@
...
+                    );
+                move_back_by_penetration_depth(
+                    collision_normal,
+                    penetration_depth,
+                    character_positioner,
+                );
+            }
+        }
+    }
+}
```
namse, line 39: doesn't it keep character seems to move to wall repeatedly even if it collide to wall but by interpolation?

luda-adventure/src/app/game/types/game_object/types/renderer.rs
```diff
@@ -0,0 +1,40 @@
...
+    }
+}
+
+pub trait RenderGameObjectList {
+    fn render(
+        &self,
+        game_state: &GameState,
+        rendering_context: &RenderingContext,
+    ) -> namui::RenderingTree;
+}
```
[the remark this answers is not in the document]
bigfoodK: That make sense. anyway it seems dangling trait. I will remove this.

luda-adventure/src/app/game/types/rendering_context.rs
```diff
@@ -0,0 +1,9 @@
+use super::Tile;
+use namui::prelude::*;
+
+pub struct RenderingContext {
+    pub current_time: Time,
```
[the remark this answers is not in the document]
bigfoodK: Definitely yes. I will fix it

luda-adventure/src/app/game/types/game_object/types/positioner.rs
```diff
@@ -0,0 +1,56 @@
...
+            previous_xy: xy,
+        }
+    }
+
+    pub fn xy(&self) -> Xy<Tile> {
+        self.current_xy
+    }
+    pub fn xy_with_interpolation(&self, interpolation_progress: f32) -> Xy<Tile> {
+        self.previous_xy * (1.0 - interpolation_progress) + self.current_xy * interpolation_progress
+    }
```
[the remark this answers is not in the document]
bigfoodK: Look good to me. Any example?

luda-adventure/src/app/game/update/resolve_collision_about_character.rs
bigfoodK: No, it won't. due to `move_back_by_penetration_depth`.
In end of tick, positioner has collision resolved xy. and `save_xy_for_interpolation` always saves collision resolved xy.
Animation

luda-adventure/src/app/game/update/calculate_next_motion_of_character/calculate_next_movement_of_character.rs
namse: That's what i'm saying. didn't we implemented that case?

luda-adventure/src/app/game/game.rs
namse: `game` just activate the function, other jobs will be done inside of job.
`game` should not control inside of the job, before they need to make a branches of logic.
the function you will call need to have the responsibility and it should hide the implementation from outside by [Information_hiding](https://en.wikipedia.org/wiki/Information_hiding).

luda-adventure/src/app/game/types/game_object/types/positioner.rs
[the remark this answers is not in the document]
namse: It shows the one-tick delayed. OK.

luda-adventure/src/app/game/update/resolve_collision_about_character.rs
```diff
@@ -0,0 +1,129 @@
...
+            x: -penetration_depth,
+            y: 0.tile(),
+        },
+        CollisionNormal::Down => Xy {
+            x: 0.tile(),
+            y: -penetration_depth,
+        },
+    };
+    character_positioner.set_xy(character_positioner.xy() + counter_penetration_xy)
+}
```
namse, line 129: It seems have bug like this video. I only press down after slightly left-moving.

https://user-images.githubusercontent.com/3580430/197388981-5d047d77-0be2-48b0-94d5-5f174e48912e.mp4

luda-adventure/src/app/game/types/game_object/types/positioner.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub fn apply_movement(&mut self, duration: Time) {
+        if let Movement::Moving(velocity) = self.movement {
+            self.current_xy.x += velocity.x * duration;
+            self.current_xy.y += velocity.y * duration;
+        }
+    }
+
+    pub fn save_xy_for_interpolation(&mut self) {
+        self.previous_xy = self.current_xy;
+    }
```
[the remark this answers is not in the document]
bigfoodK: Got it. I will merge it.

luda-adventure/src/app/game/update/resolve_collision_about_character.rs
bigfoodK: That looks annoying. I will fix it

luda-adventure/src/app/game/game.rs
bigfoodK: I intended to show the flow. I will hide this.

namse: It's enough to show how important this flow is important with function name or comment.

luda-adventure/src/app/game/update/calculate_next_motion_of_character/calculate_next_movement_of_character.rs
bigfoodK: No. We didn't

namse: `velocity * Xy::single(delta_time)` is what i'm saying

bigfoodK: Aaah I finally understood. Thankyou

bigfoodK:
I just removed poly to poly collision code.
Because it didn't working properly, and there's no scenario to do `poly to poly` yet.

luda-adventure/src/app/game/types/game_object/component/mover.rs
```diff
@@ -0,0 +1,31 @@
...
+        }
+    }
+
+    pub fn movement(&self) -> Movement {
+        self.movement
+    }
+
+    pub fn set_movement(&mut self, movement: Movement) {
+        self.movement = movement;
+    }
```
namse, line 30: You made setter and getter and they don't nothing special, so I suggest make `movement` as `pub`.

namse (changes requested):
Cool. Other code become super clean. Good job bigfood! I am really proud of you. You become really good at this code.

namse (approved):

PR #376, opened 2022-10-05
Support text_input blur

namse:
1

### DIFF
--- a/namui/src/namui/mod.rs
+++ b/namui/src/namui/mod.rs
@@ -7,7 +7,7 @@ mod namui_context;
 mod random;
 mod render;
 mod skia;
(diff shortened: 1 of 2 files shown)


namui/src/namui/mod.rs
```diff
@@ -7,7 +7,7 @@ mod namui_context;
 mod random;
 mod render;
 mod skia;
-pub(crate) mod system;
+pub mod system;
```
namse, line 10: I found that render::text_input and system::text_input conflicted so I exposed system. nothing changed but from now on user can use namui::system too.

bigfoodK (approved):

PR #378, opened 2022-10-05
Expose common_for_f32_type macro

namse:
# What happened?

When external crate tried to use `common_for_f32_type` macro of namui, there were error like below

```
error[E0119]: conflicting implementations of trait `std::ops::Mul<&f32>` for type `app::game::types::tile::Tile`
 --> src/app/game/types/tile.rs:3:1
  |
3 | namui::common_for_f32_type!(Tile, tile, TileExt);
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | |
  | first implementation here
  | conflicting implementation for `app::game::types::tile::Tile`
  |
  = note: upstream crates may add a new impl of trait `namui::Ratio` for type `&f32` in future versions
  = note: this error originates in the macro `$crate::_impl_binary_op_owned_borrowed` which comes from the expansion of the macro `namui::common_for_f32_type` (in Nightly builds, run with -Z macro-backtrace for more info)
```

The error point is here
```rust
impl<T: $crate::types::Ratio> std::ops::Mul<T> for $your_type {
    type Output = $your_type;
    fn mul(self, rhs: T) -> Self::Output {
        (self.as_f32() * rhs.as_f32()).into()
    }
}
```
That mean, 
1. we implemented `$your_type * &f32` in that macro
2. also we implemented `$your_type * Ratio` in that macro
3. So, if upstream crates(=namui) add a new impl of trait `namui::Ratio` for type `&f32`, it could be problem.

So I decided to add a impl for trait `namui::Ratio` for type `f32` and `usize`, `i32`.
Therefore, We should make operation like `type * ratio`, not `ratio * type` because we cannot make `ratio * type` override in downstream crate.

Test

image

### DIFF
--- a/namui/src/namui/common/types/macros.rs
+++ b/namui/src/namui/common/types/macros.rs
@@ -1,3 +1,4 @@
+#[macro_export]
 macro_rules! common_for_f32_type {
     ($your_type: tt, $short_term: ident, $short_term_ext: ident) => {
         $crate::types::macros::common_for_f32_type!($your_type, |lhs: $your_type| -> f32 {
@@ -50,6 +51,9 @@ macro_rules! common_for_f32_type {
                     *self
                 }
             }
+            pub fn is_finite(&self) -> bool {
+                self.0.is_finite()
+            }
         }
 
         $crate::types::macros::impl_op_forward_ref!(+|x: $your_type, y: $your_type| -> $your_type {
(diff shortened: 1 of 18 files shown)


namui/src/namui/common/types/macros.rs
```diff
@@ -50,6 +51,9 @@ macro_rules! common_for_f32_type {
                     *self
                 }
             }
+            pub fn is_finite(&self) -> bool {
+                self.0.is_finite()
+            }
```
namse, line 56: This is additional change. wink wink

namui/src/namui/common/types/rect.rs
```diff
@@ -553,3 +553,36 @@ where
...
+                bottom,
+            } => Rect::Ltrb {
+                left: left + rhs.x.clone(),
+                top: top + rhs.y.clone(),
+                right: right + rhs.x,
+                bottom: bottom + rhs.y,
+            },
+        }
+    }
+}
```
namse, line 588: This is additional change. wink wink

PR #379, opened 2022-10-08
Enhence Per type

bigfoodK:
# Features
- Impl SimpleSigned for Per
  - `is_sign_positive()` and `is_sign_negative()`.
- Impl Ord for Per


namui/src/namui/common/types/simple_signed.rs
```diff
@@ -0,0 +1,18 @@
+use num::Signed;
+
+/// It is similar to `num::Signed`,
+/// but it implements only `is_positive` and `is_negative`.
+pub trait SimpleSigned {
+    fn is_sign_positive(&self) -> bool;
+    fn is_sign_negative(&self) -> bool;
+}
```
bigfoodK, line 8: I implemented new trait `SimpleSigned` instead of `num::Signed`.
`num::Signed` requires [Num](https://docs.rs/num/latest/num/trait.Num.html) + [Neg](https://doc.rust-lang.org/nightly/core/ops/arith/trait.Neg.html)<Output = Self>.
I think that is overhead. So I made it.

Code below is about compatibility for `num::Signed`.

namse:
Please add this to `common_for_f32_type` macro too

namui/src/namui/common/types/macros.rs
```diff
@@ -142,6 +142,16 @@ macro_rules! common_for_f32_type {
                 $your_type(self as f32)
             }
         }
+
+        impl crate::SimpleSigned for $your_type {
```
namse, line 146: This should be $crate::SimpleSigned. Test it in other crate to make it sure.

namse:
You know, if you use trait instead of just impl the function, user need to import trait when they use trait's function. Do you know and think it's Ok for that inconvenience?

bigfoodK:
No, didn't know that. But I think that's ok. Because there is a prelude.

namui/src/namui/common/types/macros.rs
bigfoodK: [4dabf20](https://github.com/NamseEnt/namseent/pull/379/commits/4dabf2076dec13b783e4e8c7059937e1af8cd1aa)

namse (approved):

PR #380, opened 2022-10-11
Improve namui vector type calculating expression 

namse:
# What I did

Main purpose of this is the PR to support make simple operation for below.

```rust
// before
let xy = Xy {
  x: velocity.x * time,
  y: velocity.y * time,
};

// after
let xy = velocity * Xy::single(time);
```

- Expose vector_types macro
- Clean codes of macros
- Divide Ratio or Not Ratio case
- Replace vector_types's f32, i32 calculation to Ratio


namui/src/namui/common/types/macros.rs
```diff
@@ -1,21 +1,34 @@
...
+        }, $short_term, $short_term_ext, not_ratio);
+    };
+    ($your_type: tt, $to_f32: expr, $from: expr, $short_term: ident, $short_term_ext: ident, ratio) => {
+        $crate::types::common_for_f32_type!($your_type, $to_f32, $from, $short_term, $short_term_ext);
+    };
+    ($your_type: tt, $to_f32: expr, $from: expr, $short_term: ident, $short_term_ext: ident, not_ratio) => {
+        $crate::types::common_for_f32_type!($your_type, $to_f32, $from, $short_term, $short_term_ext);
+        $crate::types::impl_op_forward_ref!(/|x: $your_type, y: $your_type| -> f32 {
+            x.0 / y.0
+        });
```
bigfoodK, line 21: No need for Mul(*) right?

bigfoodK (approved):

namui/src/namui/common/types/macros.rs
namse: Yes it's only for same dimension division.

PR #389, opened 2022-10-30
Floor position to prevent flickering

namse:
After
https://user-images.githubusercontent.com/3580430/198870476-258aa8d4-4a5f-4b3c-8bd6-146693967deb.mp4

The reason why the character is shaking is not about floor. I think it's about the problem of interpolation with skipping ticks.
Before this commit, look at the wall, it already has been shaked.


namse:
@bigfoodK Merge this if you want, in your flavor.

PR #390, opened 2022-10-30
Fix deprecating set-output command

namse:
I got a meessage from github action
image

and they said it's deprecated.
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


bigfoodK (approved):

namse:
It seems not working in action. let me check it.

namse:
Ok it's works 
image

PR #401, opened 2022-11-02
Add text_input padding

namse:
Add padding for text_input

Tested in local with sample

https://user-images.githubusercontent.com/3580430/199461841-68d15482-4478-4dd0-b247-6883c7e34bfa.mp4

(padding was 10.px() in this test. now it become 4.px() as default)

### DIFF
--- a/namui/src/namui/render/text_input/mod.rs
+++ b/namui/src/namui/render/text_input/mod.rs
@@ -20,12 +20,11 @@ pub struct TextInput {
 #[derive(Clone, Debug)]
 pub struct Props {
     pub rect: Rect<Px>,
-    pub rect_style: RectStyle,
     pub text: String,
     pub text_align: TextAlign,
     pub text_baseline: TextBaseline,
     pub font_type: FontType,
-    pub text_style: TextStyle,
+    pub style: Style,
     pub event_handler: Option<EventHandler>,
 }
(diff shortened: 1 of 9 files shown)


namui/src/namui/render/text_input/mod.rs
```diff
@@ -112,9 +132,9 @@ impl TextInput {
             .chain(std::iter::once_with(|| get_fallback_fonts(font.size)).flatten())
             .collect::<Vec<_>>();
 
-        let paint = get_text_paint(props.text_style.color).build();
+        let paint = get_text_paint(props.style.text.color).build();
 
-        let line_texts = LineTexts::new(&props.text, &fonts, &paint, Some(props.rect.width()));
```
namse, line 117: I found the wrong code so I fixed it

bigfoodK: What was wrong?

namse: technically not wrong but it already calculated in https://github.com/NamseEnt/namseent/pull/401/files/0631087981e00ac0d9a1805400638bad93c7c62d#diff-3f651cf53eacd2fed5dad722d7b63e372818aed59498c0dbea58c681355cdb37R182 so I make it same

bigfoodK (approved):

PR #402, opened 2022-11-03
Advance text input

namse:
I think text_input is very unstable so I will make updates with PRs more.

- Support pub fn `focused_text_input_id(id)` and `last_focus_requested_text_input_id()`
- Remove `selection` from `Focus` event
- Call `Blur` and `Focus` on focus changed by Api

PR #403, opened 2022-11-03
Keep text input select on scroll

namse:
After #402

Previously text_inputs blurred when scrolled up or down so text_input is hidden.
I kept selected id and let it be focused on render stage.

local tested


namse:
Please update branch before review

bigfoodK (approved):

PR #412, opened 2022-11-07
Migrate cut

namse:
# What I did 
- Migrate rpc's data type `Cut`.
- Separate namui and namui-type to use struct in non-wasm, non-game environment(in this PR, for luda-editor/server).
- Remove `Time::now()` to separate namui-type from namui. Instead, call `namui::now()`.

# Comment
I made a migration tool, this can migrate struct using secret '_v' field in json.

Currently it's not possible to load sequence in master branch because the data structure is not same, unlike what I said in https://github.com/NamseEnt/namseent/pull/410. It makes error even user didn't use image.

So I made this migration system urgently.


luda-editor/migration/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+                                while let Some((key, value)) = tri!(visitor.next_entry()) {
+                                    values.insert(key, value);
+                                }
+
+                                Ok(Value::Object(values))
+                            }
+                            None => Ok(Value::Object(Map::new())),
+                        }
+                    }
+                }
```
namse, line 210: This code is from serde_json, to deserialize json using `deserializer.deserialize_any`.

luda-editor/migration/src/lib.rs
```diff
@@ -0,0 +1,247 @@
...
+                if version < 1 {
+                    let previous = serde_json::from_value(json_value).map_err(|e| {
+                        serde::de::Error::custom(format!("failed to deserialize previous version: {}", e))
+                    })?;
+                    Ok(#struct_name::migrate(previous))
+                } else {
+                    Ok(#struct_name {
+                        #(#fields_deserialize)*
+                    })
+                }
```
bigfoodK, line 235: Could you check this?
Is it right to use `#struct_name::migrate(previous)` when serialized version is lower than **1**?. not **current version**?
I expected the following logic.
```
if `serialized_version` is lower than `current_version`:
    do migrate
else:
    return deserialized
```

bigfoodK (changes requested):

luda-editor/migration/src/lib.rs
namse: Oh my typo. give me sec

namse: Fixed https://github.com/NamseEnt/namseent/pull/412/commits/8e3cb246ed1cf1f8652a42272e56aed93400d255

PR #434, opened 2022-11-16
Put transparent default border for text_input

namse:
text_input uses rendering tree bounding_box, the sample in namui used rect stroke. Without rect stroke, it wasn't worked as expected. I put transparent border if it doesn't have border for passed rect.

Before

https://user-images.githubusercontent.com/3580430/202150516-f2cba280-6879-42d4-bf17-7f475794eef8.mp4

After

https://user-images.githubusercontent.com/3580430/202150536-718195fc-55e5-4c70-93cd-78d126287241.mp4


namui/src/namui/render/text_input/mod.rs
```diff
@@ -144,7 +144,19 @@ impl TextInput {
         render([
             namui::rect(RectParam {
                 rect: props.rect,
-                style: props.style.rect,
+                style: RectStyle {
+                    stroke: if props.style.rect.stroke.is_some() || props.style.rect.fill.is_some()
+                    {
+                        props.style.rect.stroke
```
bigfoodK, line 150: Is `props.style.rect.fill.is_some()` related to `props.style.rect.stroke`?

namse: Yes. If fill is some, they will render rect path using fill paint, so it will be caught during calculating bounding box

bigfoodK (approved):

PR #442, opened 2022-11-16
Query components without refcell

bigfoodK:
# What I did
## Query components without refcell
Also, there is no entity returned.
### Before
```rust
let (entity: Entity, (a: Ref<A>, b: Ref<B>, c: Ref<C>)) = app.query<&A, &B, &C>();
```
### After
```rust
let (a: &A, b: &B, c: &C) = app.query<A, B, C>();
```

## Support optional component
```rust
let (a: &A, c: Option<&C>) = app.query<A, Option<C>>();
```

### DIFF
--- a/luda-adventure/src/ecs/component.rs
+++ b/luda-adventure/src/ecs/component.rs
@@ -1,0 +1,32 @@
+use super::Entity;
+use std::{cell::RefCell, fmt::Debug};
+
+pub trait Component: Debug {}
+pub trait WrappedComponent: Debug {
+    fn as_any(&self) -> &dyn std::any::Any;
+    fn as_any_mut(&mut self) -> &mut dyn std::any::Any;
+}
+impl<T: Debug + 'static> WrappedComponent for RefCell<T> {
+    fn as_any(&self) -> &dyn std::any::Any {
+        self as &dyn std::any::Any
+    }
+    fn as_any_mut(&mut self) -> &mut dyn std::any::Any {
+        self as &mut dyn std::any::Any
+    }
+}
+
+pub trait ComponentCombination<'a> {
+    type Output;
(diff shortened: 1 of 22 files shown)


namse:
> Also, there is no entity returned.

Why did you decide not to return entity?

bigfoodK:
From now on, the component belongs to the entity. So if we borrow component as mut, then we can not return entity together.

luda-adventure/src/ecs/component.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+    pub fn as_ref_mut(&mut self) -> &mut T {
+        &mut self.component
+    }
+}
+
+pub trait ContainedComponent: Debug + 'static {
+    fn as_any(&self) -> &dyn std::any::Any;
+    fn as_any_mut(&mut self) -> &mut dyn std::any::Any;
+}
```
namse, line 28: Why did you need `ComponentContainer` and `ContainedComponent`, not only just `Component`?

namse:
> From now on, the component belongs to the entity. So if we borrow component as mut, then we can not return entity together.

Then how user can determine which entity queried?

luda-adventure/src/ecs/component.rs
bigfoodK: For downcasting. If we store them as `Vec<Box<dyn component>>`, It's typeid will be `Box<dyn component>`'s, not specific component's like `Box<Mover>`.

bigfoodK:
> > From now on, the component belongs to the entity. So if we borrow component as mut, then we can not return entity together.
> 
> Then how user can determine which entity queried?
User may not know.

namse:
@bigfoodK Let's make sure what this PR want to do.
No RefCell used in last code.
What's the purpose of this PR?

bigfoodK:
Deparated

PR #443, opened 2022-11-17
Save and load ecs app

namse:
Changes 
1.  Make component hash as two dimensions(`hashmap<app.id, hashmap<entity.id, component>>`) to prevent query collision on test.
2. Save and load ecs app using ron and generated code.
3. Moved all component code into `src/component`

Someday this project needs migration system. Not now.


luda-adventure/src/component/renderer.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+}
+
+impl RenderType {
+    fn render(
+        &self,
+        entity: &crate::ecs::Entity,
+        _game_state: &GameState,
+        rendering_context: &RenderingContext,
+    ) -> RenderingTree {
```
namse, line 53: I moved all render code in here. you can edit in your flavor I think.
You mentioned that 'sprite, animation, etc...', then maybe you can put that kind of thing in here like RenderType::Sprite.

bigfoodK (approved):

PR #451, opened 2022-11-19
Render with sprites

bigfoodK:
Animation
There is nothing special.


namse:
Why is it shaking when moving?

namse (approved):

PR #454, opened 2022-11-22
Disable pixel flooring for fix character shaking

bigfoodK:
The previous issue does not appear when using sprites.
*previous issue: Unintended lines between tile.
Animation


namse:
Why it doesn't appear after apply sprite?

bigfoodK:
I don't know. I didn't check that.

PR #456, opened 2022-11-23
Hello sheet

namse:
Sheet is the new component of namui-prebuilt.

Sheet is very similar to Microsoft Excel, Google Docs Spreadsheet.

It's kind of table, but can have various height and width for cells. But every cell in same row should have same height, and every cell in same column should have same width.

I made a functionality for sheet below,
- Click event handler
- Text, Image Cell
- Copy/Paste for text cell
- text_input editing on double click the cell
- Single cell selection using one click

There are many things to do. I will put them later.

I replaced luda-editor's image-table to Sheet. It became so easy to code.

https://user-images.githubusercontent.com/3580430/203599473-5c669320-227d-4b55-8c81-08a8b4d26a7b.mp4


namse:
@bigfoodK Would you review this PR after merging #457 and #458?

bigfoodK (approved):

PR #459, opened 2022-11-23
Chain update method using Event::is() to support return

namse:
`fn namui::Event::is()` helps to make shorter and clean code for namui event handling on `update()` function.

The reason why I made this code is, previously we cannot use return statement in middle of update function because we used downcast_ref, so if we return, whole update function returned.
After using callback in `Event::is()`, we can use return to stop handling specific event type but we can keep going to next events.
User can stop whole update function using downcast_ref, `namui::Event` also support downcast_ref too.

https://namseent.zulipchat.com/#narrow/stream/332914-.EA.B8.B0.EC.88.A0.EC.97.B0.EA.B5.AC.EC.86.8C/topic/Event.20downcast.20.EA.B0.9C.EC.84.A0


namse:
@bigfoodK Please review this PR in the critical mind whether actually `Event::is` made our code better or not.

bigfoodK (approved):
LGTM
Is there anything else you're worried about?

namse:
Nothing. 
We can make a checkpoint before merge. Yes, No, Soso. 
- does this PR make code clear?
- Less bug makable?
- Add new features?
- Unnecesary changes?

bigfoodK:
- does this PR make code clear? `Yes`
- Less bug makable? `Soso`
- Add new features? `Yes`
- Unnecesary changes? `Soso`

namse:
I'm not sure why this PR merged even checks failed.

PR #462, opened 2022-11-25
Support local storage

bigfoodK:
# New Features
- `namui::file::local_storage::read(key: &str)` -> `Result<String>`
- `namui::file::local_storage::write(key: &str, content: &str)` -> `Result<()>`
- `namui::file::local_storage::delete(key: &str)` -> `Result<()>`

Unlike system_drive, it is key-value style stroage.


namse:
What's the purpose of this storage?

bigfoodK:
> What's the purpose of this storage?
> 
> Is it the `Local Persistent Storage` on this design docs? https://docs.google.com/document/d/1iBYAH94qWKdFL_pa4-HX33mKQ55USDr3dlzt5IXnFf0/edit#heading=h.uubmjht3n2jt

Yes

namse:
> > What's the purpose of this storage?
> > Is it the `Local Persistent Storage` on this design docs? https://docs.google.com/document/d/1iBYAH94qWKdFL_pa4-HX33mKQ55USDr3dlzt5IXnFf0/edit#heading=h.uubmjht3n2jt
> 
> Yes

It's not a good idea to save data as string as default. It should be byte array for compatibility.
And it uses local_storage, which is not persistent storage and has storage limit about 500MB for chrome.

You can use Cache for now because it uses indexedDB which has more quota than local storage and support binary data unlike localstorage.

I think you need to design more about local persistent storage. I think it should use Origin Private File System for web. Check below docs.
1. https://web.dev/file-system-access/#accessing-the-origin-private-file-system
2. https://webkit.org/blog/12257/the-file-system-access-api-with-origin-private-file-system/

bigfoodK:
> > > What's the purpose of this storage?
> > > Is it the `Local Persistent Storage` on this design docs? https://docs.google.com/document/d/1iBYAH94qWKdFL_pa4-HX33mKQ55USDr3dlzt5IXnFf0/edit#heading=h.uubmjht3n2jt
> > 
> > 
> > Yes
> 
> It's not a good idea to save data as string as default. It should be byte array for compatibility. And it uses local_storage, which is not persistent storage and has storage limit about 500MB for chrome.
> 
> You can use Cache for now because it uses indexedDB which has more quota than local storage and support binary data unlike localstorage.
> 
> I think you need to design more about local persistent storage. I think it should use Origin Private File System for web. Check below docs.
> 
> 1. https://web.dev/file-system-access/#accessing-the-origin-private-file-system
> 2. https://webkit.org/blog/12257/the-file-system-access-api-with-origin-private-file-system/

I will check it.

bigfoodK:
## Changes
Use `window.navigator.storage.getDirectory()` instead LocaStorage.
Tested in chrome. There is a high probability that other browsers will not support it.

### Features
- read
- write
- read_dir
- make_dir

bigfoodK:
Oh... I forgot to delete. We need that, right?

namui/src/namui/system/file/local_storage/write.rs
```diff
@@ -0,0 +1,60 @@
...
+
+    let file_stream = file_handle.create_writable().await?;
+    let writer = file_stream.get_writer()?;
+    for chunk in content.as_ref().chunks(CHUNK_SIZE) {
+        let unit8array = Uint8Array::new_with_length(chunk.len() as u32);
+        unit8array.copy_from(chunk);
+        JsFuture::from(writer.write_with_chunk(&unit8array)).await?;
+    }
+    JsFuture::from(writer.close()).await?;
+    Ok(())
```
namse, line 53: I think this code will have a performance trouble by bottle neck. Did you test the bps?
This code waits until stream write a chunk. If stream is fill with chunks, it will stop this for loop, and when stream is ready to get data after consume, we push a new chunk, so at that time, we have a delay between 'stream ready(in js)' and 'push data(wasm)', including js-wasm context switch.

First of all, do we need to chunk the data before write the data? This function gets parameter `content` which have whole data. `write_with_chunk` with chunk should be used when the source of data is stream type. But it's not a stream right now.

So, I suggest two options.
1. write whole binary data at once
2. make a stream type and get content as stream and pipe that to file's writable stream.

I think the first option is good for start. the second will be highest performance because it use full stream from source to storage. But we don't have stream yet in namui system, so we can extends this function with stream later when we make a stream type.

namse (changes requested):

namui/src/namui/system/file/local_storage/write.rs
namse: This is a good article to understand what i'm saying, https://iximiuz.com/en/posts/nodejs-readable-streams-distilled/

namse: Take a look at this to make zero-copy https://github.com/rustwasm/wasm-bindgen/issues/2456

bigfoodK: Good. As you said, there was suuuper bottleneck.
image
It was tested with only 1MiB.

## Changes
- Do not split chunks.
- Do not copy data.

namse (approved):

PR #469, opened 2022-11-29
Update cargo dependencies

bigfoodK:
There was a compatibility issue while making luda-adventure. So I updated dependencies.
The update was excuted through injecting following code to `cargo-clean-check-fmt-all.py`
```python
        print(f"cd {dir_path} && cargo update")
        exit = os.system(
            f"cd {dir_path} && cargo update")
        if exit != 0:
            print(f"\n\n-- fail cargo update on {dir_path}\n\n")
            return
```

### DIFF
--- a/image-cropper/Cargo.lock
+++ b/image-cropper/Cargo.lock
@@ -15,9 +15,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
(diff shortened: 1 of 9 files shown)


namse (approved):

namse:
I think you can upload that script for next time

PR #475, opened 2022-12-05
Return None if there is no hashmap while filtering

bigfoodK:
# What happened
Sometimes it goes panic. 
This can occur when querying components that are not registered at least one.
```rust
#set_name
    .get_or_init(|| rustc_hash::FxHashMap::default())
    .get(&app_id)
    .unwrap()
    .get(&entity.id())
```


namse (approved):
Good Catch!

PR #484, opened 2023-01-25
Load default font of typeface on init

bigfoodK:
# What?
Loads default font of FontFace during init.
*default font is a 12px font.
**12px is namui_prebuilt::typography's body font size.

# Why?
Freezing occurs when text is first drawn in an adventure. 
This occurs first once per FontFace, regardless of font size.
So preload default font on init time.

### DIFF
--- a/namui/src/namui/system/font/mod.rs
+++ b/namui/src/namui/system/font/mod.rs
@@ -52,13 +52,13 @@ pub fn with_fallbacks(font: Arc<Font>) -> Vec<Arc<Font>> {
         .chain(std::iter::once_with(|| get_fallback_fonts(font_size)).flatten())
         .collect::<Vec<_>>()
 }
-pub fn get_font_of_typeface(typeface: Arc<Typeface>, font_size: IntPx) -> Arc<Font> {
+pub fn get_font_of_typeface(typeface: &Typeface, font_size: IntPx) -> Arc<Font> {
     let key = Font::generate_id(&typeface, font_size);
     let font = FONT_SYSTEM.typeface_fonts.get(&key);
     match font {
         Some(font) => font.clone(),
         None => {
-            let font = crate_font(&typeface, font_size);
+            let font = crate_font(typeface, font_size);
             FONT_SYSTEM.typeface_fonts.insert(key, font.clone());
             font
         }
@@ -81,6 +81,6 @@ fn crate_font(typeface: &Typeface, font_size: IntPx) -> Arc<Font> {
 }
 pub fn get_fallback_fonts(font_size: IntPx) -> Vec<Arc<Font>> {
     crate::typeface::get_fallback_font_typefaces()
-        .map(|typeface| crate::font::get_font_of_typeface(typeface.clone(), font_size))
+        .map(|typeface| crate::font::get_font_of_typeface(&typeface, font_size))
         .collect()
 }
--- a/namui/src/namui/system/font/mod.rs
+++ b/namui/src/namui/system/font/mod.rs
@@ -52,13 +52,13 @@ pub fn with_fallbacks(font: Arc<Font>) -> Vec<Arc<Font>> {
         .chain(std::iter::once_with(|| get_fallback_fonts(font_size)).flatten())
         .collect::<Vec<_>>()
 }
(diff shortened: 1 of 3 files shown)


namui/src/namui/system/typeface/load_sans_typeface_of_all_languages.rs
```diff
@@ -18,6 +20,12 @@ async fn load_fallback_font_typefaces() -> Result<(), Box<dyn std::error::Error>
     let typeface = get_noto_color_emoji_typeface().await?;
```
namse, line 20: Here is typeface you need

namui/src/namui/system/typeface/load_sans_typeface_of_all_languages.rs
```diff
@@ -34,9 +42,15 @@ pub async fn load_sans_typeface_of_all_languages() -> Result<(), Box<dyn std::er
...
-        .for_each(|(typeface_type, bytes)| crate::typeface::load_typeface(&typeface_type, bytes));
+    typeface_files.iter().for_each(|(typeface_type, bytes)| {
+        crate::typeface::load_typeface(&typeface_type, bytes);
+
+        let Some(typeface) = crate::typeface::get_typeface(typeface_type.clone()) else {
+            log!("Could not load typeface {:?}", typeface_type);
+            return;
+        };
+        load_default_font_of_typeface(typeface);
+    });
```
namse, line 53: I think it's better to return typeface on `load_typeface` and use it to load default font, because you handle the case which should not happen with if-else.

namse (changes requested):

namui/src/namui/system/typeface/load_sans_typeface_of_all_languages.rs
bigfoodK: Good idea

namui/src/namui/system/typeface/load_sans_typeface_of_all_languages.rs
bigfoodK: Good idea. thx

namui/src/namui/system/typeface/load_sans_typeface_of_all_languages.rs
bigfoodK: I thought that there will be many typefaces and iter them with `get_fallback_font_typefaces`.
I think I thought too much about the future. I'll follow yours.

namse: You didn't followed. Check code again please.

namse: nvm

namui/src/namui/system/font/mod.rs
```diff
@@ -52,13 +52,13 @@ pub fn with_fallbacks(font: Arc<Font>) -> Vec<Arc<Font>> {
         .chain(std::iter::once_with(|| get_fallback_fonts(font_size)).flatten())
         .collect::<Vec<_>>()
 }
-pub fn get_font_of_typeface(typeface: Arc<Typeface>, font_size: IntPx) -> Arc<Font> {
+pub fn get_font_of_typeface(typeface: &Typeface, font_size: IntPx) -> Arc<Font> {
```
namse, line 55: I hope this function keep getting arc instead of reference. I had a problem about this before.

namse (changes requested):

namui/src/namui/system/typeface/load_sans_typeface_of_all_languages.rs
namse: I think it's not that much good idea. I think it's better to make arc or typeface and use it for loading typeface and default font.

bigfoodK: Getting simpler.

namui/src/namui/system/font/mod.rs
bigfoodK: What kind of problem did you have?

namse: The case when they want to copy reference and store it to somewhere.
`&T` cannot be stored, `Arc<T>` can be stored.

namse (approved):
Fucking cool, very namui styled code

PR #486, opened 2023-01-30
Add image url loading function

bigfoodK:
Just call `try_load_url()` repeatedly until loaded.


bigfoodK:
We can add option like adjustable `OBSERVATION_INTERVAL` and `TIMEOUT`.
Need it now?

namse (approved):

namse:
> We can add option like adjustable `OBSERVATION_INTERVAL` and `TIMEOUT`. Need it now?

It's up to you.

PR #487, opened 2023-01-30
Add username for luda-editor user

namse:
Previously, `UserDocument` didn't have username. I added it with `migration` crate which support struct version migration.
Local tested.

### DIFF
--- a/luda-editor/server/src/services/auth/impls.rs
+++ b/luda-editor/server/src/services/auth/impls.rs
@@ -51,7 +51,7 @@ impl AuthService {
             Err(error) => Err(error.to_string()),
         }
     }
-    pub async fn get_github_user_id(&self, github_access_token: String) -> Result<String, String> {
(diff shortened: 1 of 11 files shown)


luda-editor/server/src/services/auth/impls.rs
```diff
@@ -64,13 +64,17 @@ impl AuthService {
         #[derive(serde::Deserialize)]
         struct ResponseBody {
             id: u128,
+            login: String,
```
bigfoodK, line 67: I don't understand that `login` is `username`.
Is it intended?

bigfoodK (changes requested):

luda-editor/server/src/services/auth/impls.rs
bigfoodK: Ah... it's github api.

bigfoodK (approved):

PR #492, opened 2023-02-05
Load all images on start

bigfoodK:
Load all images on start.

No waiting for loading now. If It takes too long because of more images in the future, we will need a loading splash.

### DIFF
--- a/luda-adventure/src/app/game/image_loader.rs
+++ b/luda-adventure/src/app/game/image_loader.rs
@@ -33,6 +33,7 @@ impl ImageLoader {
     }
 
     fn start_load_all_images(&mut self) {
+        const CONCURRENT: usize = 4;
         let ImageLoaderState::Idle = self.image_loader_state  else{
             return;
         };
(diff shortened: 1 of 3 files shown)


luda-adventure/src/app/game/image_loader.rs
```diff
@@ -0,0 +1,115 @@
...
+    fn start_load_all_images(&mut self) {
+        const CONCURRENT: usize = 4;
+        let ImageLoaderState::Idle = self.image_loader_state  else{
+            return;
+        };
+
+        let Ok(image_urls) = get_image_urls() else {
+            namui::log!("failed to get image urls");
+            return;
+        };
```
namse, line 45: should become Failed state?

namse (changes requested):

luda-adventure/src/app/game/image_loader.rs
bigfoodK: Yes it should.

namse (approved):

PR #496, opened 2023-02-06
Fix wrong server-bin path for CI

namse:


namse:
close to test deployment

PR #504, opened 2023-02-20
Route with hash

bigfoodK:
I added `wasm-bindgen` to luda-editor for subscribing `hashchange` event.
`hashchange` event will also be fired on forward/backward button of browser clicked.
so we can move between pages using forward/backward button.
If we don't need this feature, we can remove `wasm-bindgen`.

### DIFF
--- a/luda-editor/client/src/pages/project_list_page/mod.rs
+++ b/luda-editor/client/src/pages/project_list_page/mod.rs
@@ -1,8 +1,7 @@
-use super::{sequence_list_page::SequenceListPage, *};
+use super::router::Router;
 use namui::prelude::*;
 use namui_prebuilt::*;
 use rpc::list_editable_projects::EditableProject;
-use std::sync::Arc;
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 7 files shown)


luda-editor/client/src/pages/project_list_page/mod.rs
```diff
@@ -141,9 +140,7 @@ impl ProjectListPage {
             [MouseButton::Left],
             move |event| {
                 if event.button == Some(MouseButton::Left) {
-                    namui::event::send(router::Event::Route(Arc::new(move || {
-                        router::Route::SequenceListPage(SequenceListPage::new(project_id))
-                    })));
+                    Router::move_to(format!("/sequence_list/{project_id}"));
```
namse, line 143: Could you let user don't make mistake about url? i hope user can route using enum, not string directly.

bigfoodK: sure

namse (approved):

PR #533, opened 2023-04-10
Fix wrong cache path

namse:
I read some comment that `**/cargo.lock` doesn't work for cache path, but relative path would work(`./**/cargo.lock`), so I tested and it works!


namse:
Cache miss 
image

namse:
Cache hit
image

bigfoodK (approved):

PR #534, opened 2023-04-15
Clip bounding box with transformed

bigfoodK:
I tried to use table in scroll_view. But it was shown as follows.
*Red rect is bounding box
before

After this pr. It will be shown as follows.
after

### DIFF
--- a/namui/src/namui/render/rendering_tree/bounding_box.rs
+++ b/namui/src/namui/render/rendering_tree/bounding_box.rs
@@ -76,7 +76,9 @@ impl RenderingTree {
                         .and_then(|bounding_box| {
                             let path = clip.path_builder.build();
 
-                            let clip_bounding_box = path.get_bounding_box();
+                            let clip_bounding_box = path
+                                .get_bounding_box()
+                                .map(|bounding_box| matrix.transform_rect(bounding_box));
 
                             match clip.clip_op {
                                 ClipOp::Intersect => {


namui/src/namui/render/rendering_tree/bounding_box.rs
```diff
@@ -76,7 +76,9 @@ impl RenderingTree {
                         .and_then(|bounding_box| {
                             let path = clip.path_builder.build();
 
-                            let clip_bounding_box = path.get_bounding_box();
+                            let clip_bounding_box = path
+                                .get_bounding_box()
+                                .map(|bounding_box| matrix.transform_rect(bounding_box));
```
namse, line 81: No, you should take a look inside of scroll_view code, not namui.
namui takes clip special rendering tree, which get passed path_builder and it just use path as user passed.

That mean user(in this case, scroll_view code or table or you) made a wrong clip.

namse (changes requested):

namui/src/namui/render/rendering_tree/bounding_box.rs
bigfoodK: I'll check it.

bigfoodK: I think `get_bounding_box()` was problome. When calculating the ClipNode in `get_bounding_box_with_matrix()`, the following occurs:

1. The `bounding_box` is computed for the rendering tree inside the ClipNode **with matrix applied**.
2. The computed `bounding_box` is then clipped with the Path that **without matrix applied**.


No special things found at `scroll_view` and `table`.
`scroll_view` just uses `RenderingTree.get_bounding_box()` and `table` just uses `translate` and `clip`.

namse: I got it. good

namse (approved):

PR #541, opened 2023-05-13
Reduce ci time using cache

namse:
- Build docker image, but not push it to ghcr
- ssccache added
- update cache logic of workflows
- Use Cargo.lock for all projects, including library

before: 22min (cache didn't work)
after: < 9min (if cache hit)


namse:
NOTE: I think sccache cache doesn't work for now.

bigfoodK (approved):

PR #544, opened 2023-05-18
Implement memo feature

bigfoodK:
# Snapshot
image

# Shortcuts
- [shortcut for new rpc calls](https://github.com/NamseEnt/namseent/pull/544/files#diff-a546fbc45ed186ed87cb369097c8d435b0d57501a4d3e3c4fd0458512703284bR59-R360)
  - [and implement](https://github.com/NamseEnt/namseent/pull/544/files#diff-c2af72e55d0ec930628d0a6a1133aa60b123612222e0d430d9a9674fac4974acR1)
- [shortcut for new rpc data](https://github.com/NamseEnt/namseent/pull/544/files#diff-192550043211beb68b10a4b4fb499f311eb7fe86858bf65cde14244c1e57f5a9R152-R159)


luda-editor/server/server-core/src/services/sequence/mod.rs
```diff
@@ -290,15 +293,33 @@ impl rpc::SequenceService<SessionDocument> for SequenceService {
...
                 .delete_item(SequenceDocumentDelete {
                     pk_id: req.sequence_id,
                 })
                 .delete_item(ProjectSequenceDocumentDelete {
                     pk_project_id: sequence.project_id,
                     sk_sequence_id: req.sequence_id,
+                });
+
+            let transaction = memo_ids.fold(transaction, |transaction, memo_id| {
+                transaction.delete_item(MemoDocumentDelete {
```
namse, line 316: transaction only handle 100 actions, so it would be problem.

namse (changes requested):

luda-editor/server/server-core/src/services/sequence/mod.rs
namse: I think you can delete without transaction, just delete memos first and if no fail, delete other things in transaction.

bigfoodK: I think so too. Fixed with 7550e4a

namse (approved):

PR #545, opened 2023-05-31
Fix parsing of psd to cg

bigfoodK:
- Support folder variation
- Add always-on part


bigfoodK:
> @bigfoodK Did you calculated the rect of layer? I didn't and put 'todo' on that in `import` branch.

Yes, rect of layers is calculated [here](https://github.com/NamseEnt/namseent/pull/545/files#diff-5ada8612888da3b2be910dade6ea58dce88a7b43a17338806ef7211ed5c18076R162-R186)

namse (approved):

PR #549, opened 2023-06-02
Hide always-on cg parts on cg picker

bigfoodK:
### Before
image

### After
image


namse:
@bigfoodK Please just hide always-on part on selection tab. it should be always on literally so don't have to show up to user in selection panel.

bigfoodK:
> @bigfoodK Please just hide always-on part on selection tab. it should be always on literally so don't have to show up to user in selection panel.

That makes sense

namse:
Please update title and demo on descriptions

bigfoodK:
Title and description are updated

namse (approved):

PR #550, opened 2023-06-02
Select always-on cg parts initially

bigfoodK:
Previously, we had to select always-on part variation manually.
This is nonsense. always-on variation was off initially.

After this pr, always-on variation will be selected initially.


bigfoodK:
No longer needed

PR #551, opened 2023-06-03
Enable keepalive in luda editor server

bigfoodK:
There is infinite pending problem on chrome.
This seems to be more relevant to port forwarding of vscode or ssh tunneling than keepalive.
So enable keepalive for now.


namse (approved):

bigfoodK:
Keep track #552

PR #557, opened 2023-06-26
Atom & Sequence Document Split

namse:
# 1. Atom

Atom is same with Recoil. https://recoiljs.org/

# 2. Sequence Document Split

I split sequence document's cut vector into multiple sequence cut documents and track the changes using 8-length circular index. The purpose is reduce the problem of reaching DynamoDB Document Size Limit problem. 
And also this solve the problem that updating sequence had big request body because we handled sequence in single json string, but from now on, we only send the small enum action to update sequence cut.


luda-editor/client/src/components/sequence_player.rs
```diff
@@ -38,19 +38,34 @@ impl SequencePlayer {
...
+        let project_id = project_shared_data.id();
+        if cg_files.is_none() {
+            spawn_local(async move {
+                let response = crate::RPC
+                    .list_cg_files(rpc::list_cg_files::Request { project_id })
+                    .await
+                    .unwrap();
+                namui::event::send(InternalEvent::CgFilesLoaded(response.cg_files));
+            });
+        }
```
namse, line 52: Cut Editor needs cg files to render and update them.

luda-editor/client/src/pages/sequence_edit_page/loaded/components/character_editor/render/part_picker.rs
```diff
@@ -24,7 +27,7 @@ impl CharacterEditor {
         cg_file: &CgFile,
         project_id: Uuid,
         cut_id: Uuid,
-        graphic_index: usize,
+        graphic_index: Uuid,
```
namse, line 30: It's important to keep index in Uuid because index would be different between before and after run action. I'm not that much sure about it but I think Unique index will solve a lot of problem. 
Cons is that it would be tiresome to get item using Uuid, like Random Access will be imposibble unlike Vector's usize random access.

bigfoodK: Random access is not that important at least in editor.

bigfoodK (approved):

PR #562, opened 2023-07-01
New closure type ClosurePtr<P, R> with hooks

namse:
This is the PR that changes attach_event's closure parameter to ClosurePtr<P, R> which always return true on eq().
ClosurePtr keeps the pipe to closure and when user change the endpoint-closure, it also changes the target to invoke automatically. It's kind of linked-list.

The purpose of ClosurePtr is to prevent re-rendering when event handler changed.

I added the hooks sample which takes the big benefit of ClosurePtr's always-return-true-on-eq, because hooks compare props when it re-render the component.

Because of changes and dumb rustc, sometime we need to put type annotation for closure. It's easy to put on vscode using `ctrl + .` so I don't worry about it.


namui/sample/hooks/.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "rust-analyzer.linkedProjects": ["./Cargo.toml", "./Cargo.toml"]
+}
```
bigfoodK, line 3: What does it means?

namse: I... have no idea

namui/sample/hooks/rust-toolchain.toml
```diff
@@ -0,0 +1,2 @@
+[toolchain]
+channel = "nightly"
```
bigfoodK, line 2: Hooks are available only in nightly?

namse: No, it's not.

namui/sample/hooks/.vscode/settings.json
namse: I removed this.

namui/sample/hooks/rust-toolchain.toml
namse: Removed

bigfoodK (approved):

PR #563, opened 2023-07-01
Use proc-macro for rpc

namse:
I changed rpc macro to proc-macro. not that much reason to do this.

# Changes to rpc
1. no `pub` visibility for `struct Request` and  `struct Response`
2. should put `enum` in front of `Error`


bigfoodK (approved):

bigfoodK (approved):

bigfoodK:
I just confused
> no pub visibility for struct Request and struct Response

now ok

PR #570, opened 2023-07-11
Fix rgba_idx to return none for out-ranged pixel

bigfoodK:
image
- Red box is layer rect
- Blue box is psd rect

If we tried to get index of `I1`, It should be None, not index of `E1`.


namse (approved):
LGFM

PR #573, opened 2023-07-15
Expose image default shader

bigfoodK:


namse:
What's the purpose for exposing default shader?

bigfoodK:
Soon, I will use image shader for custom image rendering.
The `make_shader()` is public, but I don't need a new shader, I just need a default shader without creating everytime.

namse (approved):

PR #574, opened 2023-07-15
Update makeShaderWithChildren function of canvaskit

bigfoodK:
`@param isOpaque` seems to be deprecated.


namse (approved):
Confirmed 
https://unpkg.com/browse/canvaskit-wasm@0.38.0/types/index.d.ts#L2714

PR #583, opened 2023-07-17
Release and publish in single workflow

namse:
Github actions doesn't call another github actions in single github-token so this fucking workflows never invoke push:tag. si bal


namse:
Tested
- Tag creation
- Release creation
- if conditions

bigfoodK (approved):

PR #587, opened 2023-07-27
Hooks-v2 on namui samples

namse:
I made hooks-v2 as namui samples.

Please check README.md before review.

I will put it to namui later with testing luda-editor porting.


namui/sample/hooks-v2/src/hooks/v2/hooks/hooks/state.rs
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+
+    let state: &State = state_list[state_index]
+        .as_ref()
+        .as_any()
+        .downcast_ref()
+        .unwrap();
+
+    let state: &State = unsafe { std::mem::transmute(state) };
```
namse, line 49: Found memory vulnerability here.
If instance dead, this reference would be point very wrong memory.

namse: It's not problem for `rendering_tree`'s `attach_event` because `rendering_tree` will be re-render right after update of component tree.
But the problem is `async` code. currently `Sig` is `'static` so it could be copy into `async move` statement and it will point wrong memory if the component is dead, but the async function is working yet.

namse: To solve this problem, we need to bind `use_state`'s lifetime, to do that, we need to put ctx as parameter.

PR #625, opened 2023-09-01
Fix DeleteGraphic action to keep graphic order

bigfoodK:
I used `swap_remove()` on `cut.screen_graphics` because I just heard It is faster than `remove()`.
But `swap_remove()` changes `cut.screen_graphics`'s order.


namse (approved):
I also suspected about swap_remove on previous pr. yes, swap_remove bring last item to position.

namse:
This vec is enoughly small so you don't have to think about performance

PR #666, opened 2023-09-09
Add image picker to add uploaded image to cut

### LINKED ISSUE #659: Add image button in context like `Add cg` button

bigfoodK:
- Add `Add Image` button to cut_editor's context menu
image

- Add image_picker
image

- Fix minor bug. image_picker and cg_picker are not close on click cut_editor. It should be closed because cut_editor is out side of picker
Animation

Closes #659


luda-editor/client/src/pages/sequence_edit_page/components/image_picker/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+        let (images, _set_images) = ctx.atom(&IMAGES_ATOM);
+
+        let on_internal_event = |event: InternalEvent| match event {
+            InternalEvent::ImageThumbnailClicked { image_id } => {
+                add_image_to_cut(image_id, cut.id);
+                on_event(Event::Close);
+            }
+        };
+
+        let max_items_per_row = ctx.memo(|| (wh.width / (THUMBNAIL_WH.width)).floor() as usize);
```
namse, line 49: This must be trouble because this memo doesn't have any signal so it won't update when wh is changed.

luda-editor/client/src/pages/sequence_edit_page/components/image_picker/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+            })(wh, ctx)
+        });
+
+        let background = simple_rect(wh, color::STROKE_NORMAL, 1.px(), color::BACKGROUND)
+            .attach_event(|event| {
+                if let namui::Event::MouseDown { event } = event {
+                    event.stop_propagation();
+                    if !event.is_local_xy_in() {
+                        on_event(Event::Close);
+                    }
```
namse, line 75: Why it closes the picker?

namse (changes requested):

luda-editor/client/src/pages/sequence_edit_page/components/image_picker/mod.rs
bigfoodK: Ah, true

luda-editor/client/src/pages/sequence_edit_page/components/image_picker/mod.rs
bigfoodK: Clicking outside of picker means `User wants to do something other than adding image`.
So if user clicks outside, picker should be closed.

namse: Then all windows on right side will be closed when click left side?

namse: So, it's the problem that 'adding cg' and 'adding image' is showing on right panel which user don't expect it is temporary input panel, unlike modal.

bigfoodK: Then we should use modal design for pickers. right?

bigfoodK: > Then we should use modal design for pickers. right?

If it is right, I think it would be good to merge this pr and create a new issue.

namse: modal isn't the only option, i guess. it should be considered in other options after studying other app's cases.

namse (approved):

namse:
@bigfoodK Please create and track issue about https://github.com/NamseEnt/namseent/pull/666#discussion_r1320587941

PR #668, opened 2023-09-09
Divide clippy and clippy-fix

namse:
1. clippy --fix doesn't throw error even it fail to fix
2. they said completion script will work if it placed to ~/.local/share/bash-completion/completions

### DIFF
--- a/.github/workflows/all.yml
+++ b/.github/workflows/all.yml
@@ -50,7 +50,7 @@ jobs:
           key: ${{ runner.os }}-${{ env.CARGO_VERSION }}-command-${{ matrix.dir }}-${{ hashFiles(env.CARGO_LOCK_PATH) }}
 
       - run: |
-          for-all-projects run --only ${{ matrix.dir }} --check --fmt --fix --clippy --test
+          for-all-projects run --only ${{ matrix.dir }} --check --fmt --clippy --test
 
           if [[ $(git diff --name-only) ]]; then
               echo "::error::File change detected. Please check."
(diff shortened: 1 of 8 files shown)


.github/workflows/all.yml
```diff
@@ -50,7 +50,7 @@ jobs:
           key: ${{ runner.os }}-${{ env.CARGO_VERSION }}-command-${{ matrix.dir }}-${{ hashFiles(env.CARGO_LOCK_PATH) }}
 
       - run: |
-          for-all-projects run --only ${{ matrix.dir }} --check --fmt --fix --clippy --test
+          for-all-projects run --only ${{ matrix.dir }} --check --fmt --clippy --test
```
namse, line 53: no need to run `--fix` because `--check` will catch that case

for-all-projects/build.rs
```diff
@@ -1,22 +1,26 @@
...
+    let outdir = PathBuf::from(home).join(".local/share/bash-completion/completions");
+    let bin_name = "for-all-projects";
 
     create_dir_all(&outdir)?;
 
     let mut cmd = Cli::command();
-    generate_to(Bash, &mut cmd, "for-all-projects", outdir)?;
+    generate_to(Bash, &mut cmd, bin_name, &outdir)?;
 
     Ok(())
```
namse, line 25: You can do this later for namui-cli.

bigfoodK: Got it. #669

bigfoodK (approved):

PR #676, opened 2023-09-12
Fix ghost render - bounding box change

### LINKED ISSUE #636: [Bug]: Strange void after resize
### Expected Behavior

Button should render well

### Actual Behavior

But it was hidden by strange black rectangle. 

https://github.com/NamseEnt/namseent/assets/3580430/7bc37e30-693d-4752-a0be-c7a8c661ee7f

### Steps to Reproduce

1. start luda-editor with small window
2. after loading project list page, resize the window
3. project button is hidden by something.

### Other Info

_No response_

namse:
Closes #636 

Concept of ghost rendering is little bit difficult.
I changed it to get option to handle event or not.

### DIFF
--- a/namui/src/namui/hooks/ctx/render_ctx/public.rs
+++ b/namui/src/namui/hooks/ctx/render_ctx/public.rs
@@ -51,16 +51,6 @@ impl<'a> RenderCtx {
         }
     }
 
-    pub fn ghost_render(&self, component: impl Component) -> RenderingTree {
-        self.disable_event_handling();
-        let rendering_tree = self.render_children(
-            KeyVec::new_child(self.get_next_component_index()),
-            component,
-        );
-        self.enable_event_handling();
-        rendering_tree
-    }
-
     pub fn done(&self) -> RenderDone {
         let vec: Vec<RenderingTree> = std::mem::take(self.children.lock().unwrap().as_mut());
         let rendering_tree = crate::render(vec);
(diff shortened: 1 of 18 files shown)


namui/src/namui/hooks/ctx/render_ctx/public.rs
```diff
@@ -73,23 +63,78 @@ impl<'a> RenderCtx {
...
+                enable_event_handling: true,
+            },
+        );
+        self.children.lock().unwrap().push(rendering_tree);
+        self
+    }
+}
+
+pub struct GhostComposeOption {
+    pub enable_event_handling: bool,
```
bigfoodK, line 139: What is this used for?
It's set up, but it doesn't seem to be used anywhere

namse: It used so clippy or cargo check doesn't catch that.

https://github.com/NamseEnt/namseent/pull/676/files#diff-72d3f427136795e439949f3f0a84ef2ea33d0b903b21388799c90d20ebd7f088R166

bigfoodK: Yes, the variable `enable_event_handling` is used.
But it is only for setting value. no one actually use this.


```rust
        let prev_enable_event = self.tree_ctx.enable_event_handling(enable_event_handling);

        let rendering_tree =
            self.renderer
                .render(key_vec, component, self.matrix, self.clippings.clone());

        self.tree_ctx.enable_event_handling(prev_enable_event);
```
Above is setting `enable_event_handling` to true, and render, then restore it.
But inside rendering, `enable_event_handling` is not used.

I guess the code that prevents event_handling seems to be missing according to the `enable_event_handling` value.
Can you check?

namse: You mean this? https://github.com/NamseEnt/namseent/pull/676/files#diff-4989163211ebf99f560b38bf8b1c0a34a2ca766e63aa3b8a297b01cd9e9667f2L33

Right

namse: Fixed

bigfoodK (approved):

PR #689, opened 2023-09-18
Prevent change order to after itself

bigfoodK:
Order changing functions works like
1. Remove self
2. Insert self after `after_*`

If `after_*` is indicating self, Inserting after `after_*` is impossible because self is already removed.
So ignore order change request if `after_*` is indicating self.

### DIFF
--- a/luda-editor/client/src/pages/sequence_edit_page/sequence/mod.rs
+++ b/luda-editor/client/src/pages/sequence_edit_page/sequence/mod.rs
@@ -65,10 +65,9 @@ impl SequenceWrapped {
                         sequence.cuts.swap_remove(position);
                     }
                 }
-                rpc::data::SequenceUpdateAction::MoveCut {
-                    cut_id,
-                    after_cut_id,
-                } => {
+                rpc::data::SequenceUpdateAction::MoveCut(move_cut_action) => {
+                    let cut_id = move_cut_action.cut_id();
+                    let after_cut_id = move_cut_action.after_cut_id();
                     let moving_cut_position = sequence
                         .cuts
                         .iter()
--- a/luda-editor/client/src/pages/sequence_edit_page/sequence/mod.rs
+++ b/luda-editor/client/src/pages/sequence_edit_page/sequence/mod.rs
@@ -68,27 +68,30 @@ impl SequenceWrapped {
                 rpc::data::SequenceUpdateAction::MoveCut {
                     cut_id,
                     after_cut_id,
-                } => {
-                    let moving_cut_position = sequence
-                        .cuts
-                        .iter()
-                        .position(|cut| cut.id == cut_id)
-                        .unwrap();
-                    let moving_cut = sequence.cuts.remove(moving_cut_position);
-                    let insert_position = match after_cut_id {
-                        Some(after_cut_id) => {
-                            let position = sequence
-                                .cuts
-                                .iter()
-                                .position(|cut| cut.id == after_cut_id)
-                                .unwrap();
-                            position + 1
-                        }
-                        None => 0,
-                    };
+                } => match after_cut_id {
+                    Some(after_cut_id) if after_cut_id == cut_id => {}
+                    after_cut_id => {
+                        let moving_cut_position = sequence
+                            .cuts
+                            .iter()
+                            .position(|cut| cut.id == cut_id)
+                            .unwrap();
+                        let moving_cut = sequence.cuts.remove(moving_cut_position);
+                        let insert_position = match after_cut_id {
+                            Some(after_cut_id) => {
+                                let position = sequence
+                                    .cuts
+                                    .iter()
+                                    .position(|cut| cut.id == after_cut_id)
+                                    .unwrap();
+                                position + 1
+                            }
+                            None => 0,
+                        };
 
-                    sequence.cuts.insert(insert_position, moving_cut);
-                }
+                        sequence.cuts.insert(insert_position, moving_cut);
+                    }
+                },
                 rpc::data::SequenceUpdateAction::SplitCutText {
                     cut_id,
                     new_cut_id,
--- a/luda-editor/client/src/pages/sequence_edit_page/sequence/mod.rs
+++ b/luda-editor/client/src/pages/sequence_edit_page/sequence/mod.rs
@@ -3,7 +3,7 @@ mod syncer;
 
 use self::{history::*, syncer::*};
 use namui::Uuid;
-use rpc::data::{CutUpdateAction, Sequence};
+use rpc::data::{CutUpdateAction, MoveCutAction, Sequence};
 
 const MAX_EDIT_HISTORY: usize = 8;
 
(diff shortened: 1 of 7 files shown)


namse (changes requested):

bigfoodK:
> Hey, it should not send a request when it doesn't change order, is it make sense?

client should not send a request when it doesn't change order. 
But it would be nice if server can hanlde it naively.

luda-editor/client/src/pages/sequence_edit_page/sequence/mod.rs
```diff
@@ -68,27 +68,30 @@ impl SequenceWrapped {
...
+                                position + 1
+                            }
+                            None => 0,
+                        };
 
-                    sequence.cuts.insert(insert_position, moving_cut);
-                }
+                        sequence.cuts.insert(insert_position, moving_cut);
+                    }
+                },
```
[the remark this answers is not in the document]
bigfoodK: Fixed with efd1cad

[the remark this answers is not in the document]
bigfoodK: And I didn't know that I can use label and break in match.

namse:
> > Hey, it should not send a request when it doesn't change order, is it make sense?
> 
> client should not send a request when it doesn't change order. But it would be nice if server can hanlde it naively.

Why? That mean client sends meaningless request to server.

bigfoodK:
> > > Hey, it should not send a request when it doesn't change order, is it make sense?
> > 
> > 
> > client should not send a request when it doesn't change order. But it would be nice if server can hanlde it naively.
> 
> Why? That mean client sends meaningless request to server.

Because I think that even if a client sends a meaningless request, server should not cause an error.
Changing order with self is meaningless but could happen.
So I think it's better to ignore meaningless  request rather than throw an error.

namse:
Without error, nobody can catch that leak.
Makes error as early as you can, so make a service better.
No error is best, but hiding error is not the best.

As I wrote the comment on that source code, I planed to merge update to reduce meaningless update.

bigfoodK:
I got it. Let's throw error on that case.

namse:
> I got it. Let's throw error on that case.

Basically, Server stacks history for undo and redo. it should be exactly same with client.
So, If client decided not to stack the history, client must not send the request.

luda-editor/client/src/pages/sequence_edit_page/sequence/mod.rs
```diff
@@ -69,6 +69,9 @@ impl SequenceWrapped {
                     cut_id,
                     after_cut_id,
                 } => {
+                    if after_cut_id == Some(cut_id) {
+                        panic!("Cannot move cut after itself");
+                    }
```
namse, line 74: You didn't get what i'm saying.
You don't have to make a panic when it can be filtered.
Do filter at this line, so stop sending meaningless request to client and server.

Why did you make a panic here?

namse (changes requested):

luda-editor/client/src/pages/sequence_edit_page/sequence/mod.rs
namse: It seems blocked by this code.
https://github.com/NamseEnt/namseent/blob/8cb4c8a37ec98a5a8bf9f80836dc1a4898afdefa/luda-editor/client/src/pages/sequence_edit_page/components/cut_list_view/mod.rs#L160-L171

Why did you need this kind of panic? for what case?

bigfoodK: Developer may accidentally forget to filter that case while developing client someday.
At that time, the server should still check the case so that the developer can notice the error as soon as possible.

namse: I don't get it. 
Just make assert than, not if-and-panic.
And this PR title and description is not about this assertion.

Why do you want to this? what's the purpose? Preventing future error?

Don't block server. just block client. It's very logically same that server code could be changed so it accidently forget to filtering this case.

bigfoodK: Purpose of this pr was ignoring this edge case that causing below problem.
> If after_* is indicating self, Inserting after after_* is impossible because self is already removed.
So ignore order change request if after_* is indicating self.

And purpose is changed to preventing future error with assertion after your comment below.
> Without error, nobody can catch that leak.
Makes error as early as you can, so make a service better.
No error is best, but hiding error is not the best.

bigfoodK: > Don't block server. just block client.

Why we should filter this case at server or client one-side?
Isn't it right to filter both side because we can make mistakes on both sides?

namse: Both side could be changed at the same time. Adding `assert!` doesn't solve the problem.

Hey, then why you make the client, instead of remote mirroring like xbox or steam cloud play?

We can believe client because even if user abuse client, it doesn't affect to other project and our business.
So you don't have to make a complex server which react to everything about client-side happening.
Server only save data in database, that's the only different thing about cut editing against client.
The storage layer is different between server and client so that's why we have to implement Action's handler in server too, not for security or bug-catch reason.

So, you said the purpose is preventing index problem.
Then make a assertion at client side if you need.
Do we need? actually I don't have any idea about it. Why need it?
If you really need it, make a data structure which prevent this problem in general. Then it super force programmer don't make that kind of coding error.

namse: Don't make program complex. We are the only people who must maintain it.
I want simple and well-maintainable code.

No 3-place assertion. I very disagree with current PR's suggestion.

bigfoodK: I got a good idea
How about adding a test for that case?

bigfoodK: Nah. nevermind.

bigfoodK:
@namse Can I close this pr?

luda-editor/rpc/src/data/cut_update_action.rs
```diff
@@ -53,10 +54,36 @@ pub enum CutUpdateAction {
     DeleteGraphic {
         graphic_index: Uuid,
     },
-    ChangeGraphicOrder {
+    ChangeGraphicOrder(ChangeGraphicOrderAction),
+}
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct ChangeGraphicOrderAction {
+    pub(crate) graphic_index: Uuid,
+    pub(crate) after_graphic_index: Option<Uuid>,
```
namse, line 62: If this is opened, they can create it directly.
Put private phantom to force them use `new` function.

namse: Ah, it's rpc crate!
Ok, my suggestion is pub graphic_index and after_graphic_index, but put phantom to force them use new function for creation.

bigfoodK: Good suggestion.
I considered following three methods

1. Hide whole struct and expose constructor 
1. Using phantom as you said
1. Using `#[non_exhaustive]`

I thought that `2.` and `3.` are bit of a hack. It would be difficult to someone read this code.
On the other hand, I think `1.` is a very basic method. I think anyone can easily understand this.
So I used `1.`.

But now I realized. phatom_data is generally used for that kind of purpose. Let me try phantom.

namse (approved):
Personally I think this PR is trying to do over-engineering. I don't know whether it will prevent future error or not, so I'm gonna stop thinking about value of this PR.

namse:
And one more thing, I hope you spend more time about making system first, then add this kind of error checking.
I want editor first.

PR #693, opened 2023-09-19
Split cut text with shortcut

### LINKED ISSUE #688: Shortcut for split cut text
On some keybind, copy screen graphics and name to right next cut.
And split cut text with text cursor.

Text before cursor should remain previous cut,
And Text after cursor should be moved to newly created next cut.

bigfoodK:
Animation

Closes #688


namse (approved):
LGTM

PR #710, opened 2023-10-27
Add user acl management page

### LINKED ISSUE #407: [Luda editor] Put user editor permission
I want to give editor permission to other user update image instead of me.

bigfoodK:
We need better solution for adding user to project editor. #711

## Features
- Manage acl
  - image
- Copy user id
  - image
- List editor granted project also
  - Previously, only projects owned were listed
- `update or create` command in db.

Closes #407

### DIFF
--- a/luda-editor/server/server-core/src/apis/project/edit_user_acl.rs
+++ b/luda-editor/server/server-core/src/apis/project/edit_user_acl.rs
@@ -33,15 +33,35 @@ pub async fn edit_user_acl(
     match permission {
         Some(permission) => crate::dynamo_db()
             .transact()
-            .put_item(UserInProjectAclDocument {
-                project_id,
-                user_id,
-                permission,
+            .update_or_create_item(UserInProjectAclDocumentUpdateOrCreate {
+                pk_project_id: project_id,
+                sk_user_id: user_id,
+                update: move |mut document| async move {
+                    document.permission = permission;
+                    Ok(document)
+                },
+                create: move || async move {
+                    Ok(UserInProjectAclDocument {
+                        user_id,
+                        project_id,
+                        permission,
+                    })
+                },
             })
-            .put_item(ProjectAclUserInDocument {
-                user_id,
-                project_id,
-                permission,
+            .update_or_create_item(ProjectAclUserInDocumentUpdateOrCreate {
+                pk_user_id: user_id,
+                sk_project_id: project_id,
+                update: move |mut document| async move {
+                    document.permission = permission;
+                    Ok(document)
+                },
+                create: move || async move {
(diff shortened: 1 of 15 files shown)


luda-editor/server/server-core/src/apis/project/edit_user_acl.rs
```diff
@@ -33,21 +33,15 @@ pub async fn edit_user_acl(
...
-                pk_user_id: session.user_id,
-                sk_project_id: project.id,
-                update: move |mut document: ProjectAclUserInDocument| async move {
-                    document.permission = permission;
-                    Ok(document)
-                },
+            .put_item(ProjectAclUserInDocument {
+                user_id,
+                project_id,
+                permission,
```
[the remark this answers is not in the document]
namse: No, it will be big problem because put-item doesn't make a lock.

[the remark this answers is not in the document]
bigfoodK: Changed with a501565
Added the update or create command.

luda-editor/server/server-core/src/apis/project/edit_user_acl.rs
```diff
@@ -33,21 +33,35 @@ pub async fn edit_user_acl(
...
-                update: move |mut document: ProjectAclUserInDocument| async move {
+            .update_or_create_item(ProjectAclUserInDocumentUpdateOrCreate {
+                pk_user_id: user_id,
+                sk_project_id: project_id,
+                update: move |mut document| async move {
                     document.permission = permission;
                     Ok(document)
                 },
+                create: move || async move {
+                    Ok(ProjectAclUserInDocument {
```
namse, line 59: Other thing seems really good but would you remove Result type for create part? 
Update needs Result because it should check saved data after get item. 
But create doesn't.

namse (changes requested):

luda-editor/server/server-core/src/apis/project/edit_user_acl.rs
namse: Hmm... wait. maybe future + Result for create would be make sence... right? how do you think?

bigfoodK: I don't know. I just used this for flexibility.

I thought `Create` could be async and maybe fail, not just the synchronous constructor.
But I didn't find this case yet.

So I think it's also okay to remove result type for `create`

namse: Let it be Future<Result>. I think it's right.

namse (approved):
LGTM
But honestly I didn't read code very much, detailly wwwwwww

PR #716, opened 2023-11-05
Blur text area element on TextInput blured

### LINKED ISSUE #702: [Bug]: Key down event doesn't occur after using text_input
### Expected Behavior

Key down event on `Component.attach_event` component should work.

### Actual Behavior

No key down event occurs in `Component.attach_event`.

### Steps to Reproduce

1. Use text_input
image
2. Test key down event handler in `Component.attach_event`.
image
In this case, moving cg with arrow key doesn't work after using text_input.

### Other Info

_No response_

bigfoodK:
Closes #702


bigfoodK:
Open it later

PR #734, opened 2023-12-27
Update rust-analyzer command for build script

namse:
Build script has different command system so I override it too


bigfoodK (approved):
It works for me

PR #750, opened 2024-01-01
Divide main thread and tokio runtime

### LINKED ISSUE #748: [Bug]: State initializes abnormally
### Expected Behavior

Don't initialize even though `set_state()` called.

### Actual Behavior

Initialization function called abnormally.

### Steps to Reproduce

1. Go to b47748ad4ee2b91c567835c309d9b0ca21a2879b
1. Modify render function of `Player` in `note-register/src/app/player/mod.rs`.
    ```rust
    impl Component for Player<'_> {
        fn render(self, ctx: &namui::prelude::RenderCtx) -> namui::prelude::RenderDone {
            let Self { wh, loaded } = self;
    
            const BUTTON_HEIGHT: Px = px(64.0);
    
            let (start_offset_ms, set_start_offset) = ctx.state(|| {
                namui::log!("init");
                0.0
            });
    
            ctx.compose(|ctx| {
                vertical([fixed(BUTTON_HEIGHT, |wh, ctx| {
                    ctx.add(Slider {
                        wh,
                        value: *start_offset_ms,
                        min: 0.sec().as_secs_f32(),
                        max: 227.sec().as_secs_f32(),
                        on_change: &|value| {
                            set_start_offset.set(value);
                            namui::log!("set: {value}");
                        },
                    });
                })])(wh, ctx);
            });
    
            ctx.done()
        }
    }
    ```
1. Drag slider
then you can see like this
image

### Other Info

- Not sure if it's a bug in namui.

namse:
Our hooks system is made on the single threaded idea but previous code didn't make sure it was on same thread.
I changed it to run hooks on single thread.

It still needs optimization for drawing, because it runs skia rendering everytime, it should be vsync.
vsync doesn't mean that user input will be ignored, just drawing on monitor will be ignored between monitor frame timing.

~That optimization will be on next pr.~  https://github.com/NamseEnt/namseent/pull/750/commits/602044b8578c3036c3cc9c22b2bdd543b3c7ea25 is the optimization.

Tested on `mouse` sample, only `init of state` was called once

closes #748


namse:
closes 748

namse:
bug occurs. effect doesn't work.

namse (commented):
10ms refresh event loop

namse:
>  bug occurs. namui spawn doesn't work.
Fixed

bigfoodK (approved):

PR #763, opened 2024-01-03
Turn on target feature based on steam hw survey

namse:
Previously I turned on target feature in wrong way so it doesn't worked when we do wasm platform.

I turned on target features, which supported by 99% of steam users' cpu.

Only once tested.

tested code
```rust
fn consume(&mut self, output: &mut [f32]) {
    self.buffer
        .drain(..output.len())
        .zip(output)
        .for_each(|(src, dst)| {
            *dst += src;
        });
}
```

### Debug, no target feature

avg: 45.824µs

### Debug, with target feature

avg: 49.48µs // what!? but it's possible because I only tested once.

### Release without target feature

avg: 8.42µs

### Release with target feature

avg: 7.828µs

### DIFF
--- a/namui/namui-cli/src/services/runtime_project/wasm.rs
+++ b/namui/namui-cli/src/services/runtime_project/wasm.rs
@@ -1,14 +1,13 @@
 use super::{get_project_name, GenerateRuntimeProjectArgs};
-use crate::*;
+use crate::{util::recreate_dir_all, *};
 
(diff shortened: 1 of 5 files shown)


namse:
Please report if you have a slower compile time after this pr.

bigfoodK (approved):
I will report if I have a slower compile time after this pr

PR #805, opened 2024-02-16
Clippy on `namui-type`

bigfoodK:
https://rust-lang.github.io/rust-clippy/master/index.html#/blocks_in_conditions


namse (approved):
I changed name `cached` to `result`. It's not just a cache.

PR #833, opened 2024-03-29
Fix wrong bounding box check

namse:
Not sure it can cover all case.


namse:
Relevant https://github.com/NamseEnt/namseent/pull/1081

Issue #870, opened 2024-06-22
[Bug]: Can not load bundled image in wasi env


bigfoodK:
### Expected Behavior

Image should be loaded same as in case of `x86_win`.

### Actual Behavior

```text
[WASI stdout] Image not loaded
[WASI stdout] Image load error: Unknown(reqwest::Error { kind: Builder, url: Url { scheme: "bundle", cannot_be_a_base: true, username: "", password: None, host: None, port: None, path: "background.jpg", query: None, fragment: None }, source: BadScheme })
```

`ctx.image()` uses `namui::network`.
In wasi, we may need to use `std::fs::read` instead of `namui::network` for bundle scheme.

### Steps to Reproduce

```rust
fn render(ctx: &RenderCtx) {
    const BACKGROUND: &str = "bundle:background.jpg";

    let background = ctx.image(BACKGROUND);
    ctx.effect("log image load status", || {
        let Some(image) = background.deref() else {
            namui::log!("Image not loaded");
            return;
        };

        match image {
            Ok(_image) => {
                namui::log!("Image loaded");
            }
            Err(error) => {
                namui::log!("Image load error: {error}");
            }
        }
    });
}
```

PR #872, opened 2024-06-24
Use `ResourceLocation` for image

### LINKED ISSUE #871: Replace url to ResourceLocation
Example
```
ResourceLocation {
  Bundle(Path),
  LocalStroage(Path),
  Http(Url),
}
```

bigfoodK:
It's only applied to image. Audio may will be next time we need.

Closes #871 #870

### DIFF
--- a/namui/sample/image-load-background/src/lib.rs
+++ b/namui/sample/image-load-background/src/lib.rs
@@ -9,52 +9,77 @@ pub fn main() {
 struct App;
 
 impl Component for App {
-    fn render(self, ctx: &RenderCtx)  {
+    fn render(self, ctx: &RenderCtx) {
         let size = namui::screen::size();
 
         // let jpg_length = 14;
         let jpg_length = 1;
         let png_length = 10;
-        let jpgs = (0..jpg_length)
-            .map(|index| Url::parse(&format!("bundle:resources/{index}.jpg")).unwrap());
-        let pngs = (0..png_length)
-            .map(|index| Url::parse(&format!("bundle:resources/{index}.png")).unwrap());
+        let jpgs =
+            (0..jpg_length).map(|index| ResourceLocation::bundle(format!("resources/{index}.jpg")));
+        let pngs =
+            (0..png_length).map(|index| ResourceLocation::bundle(format!("resources/{index}.png")));
 
-        let image_urls = jpgs.chain(pngs).collect::<Vec<_>>();
+        let image_resource_locations = jpgs.chain(pngs).collect::<Vec<_>>();
 
         let x_index_length = 6;
-        let y_index_length = (image_urls.len() as f32 / x_index_length as f32).ceil() as usize;
+        let y_index_length =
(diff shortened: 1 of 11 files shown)


namui/sample/image-load-background/src/lib.rs
```diff
@@ -9,52 +9,77 @@ pub fn main() {
...
+        let y_index_length =
+            (image_resource_locations.len() as f32 / x_index_length as f32).ceil() as usize;
         let image_width = size.width.into_px() / x_index_length as f32;
         let image_height = size.height.into_px() / y_index_length as f32;
 
         ctx.compose(|ctx| {
             for x in 0..x_index_length {
                 for y in 0..y_index_length {
-                    if let Some(image_url) = image_urls.get(x + y * 6) {
+                    ctx.compose(|ctx| {
```
namse, line 34: Should we need to use ctx.compose to use image?

namui/sample/image-load-background/src/lib.rs
```diff
@@ -9,52 +9,77 @@ pub fn main() {
...
+            };
+            ctx.add(namui::image(ImageParam {
+                rect,
+                image: image.clone(),
+                style: ImageStyle {
+                    fit: ImageFit::Contain,
+                    paint: None,
+                },
+            }));
+        });
```
namse, line 83: I think this should be the default behavior supported by namui.
Most of cases, user don't want to take care whether the image is loaded or not, and they just want to put src for namui to draw image.

Please consider below cases
1. Manually load image, get image info, then render.
2. Just put image source to namui, Don't care whether image is loaded or not exists, like web's img.

The second option may break the perfectness, But I don't want user to check image's loading and existence every time, especially when the image is from bundle, which we can make sure it should be loaded.

namse (changes requested):

namui/sample/image-load-background/src/lib.rs
bigfoodK: Not really

namui/sample/image-load-background/src/lib.rs
bigfoodK: Then we can do the followings

1. Keep `ImageDrawCommand` for case 1
2. Change `namui::image()` to receive just source for case 2

namse: I guess there is `ImageRender` https://github.com/NamseEnt/namseent/pull/872/files#diff-b4dc7e7c2474ffadea70322783d4d25589cdfc21a17526dca895db2ddb11b3f8R31 so I don't think we need to provide another wrapper?

namse: If ImageRender can cover that case, could you use it and make the sample simple? Because sample is the kind of manual of namui, so I hope it should be simple as we want it to be.

bigfoodK: I didn't know that there is `ImageRender`
sample updated

namse (approved):
lgtm

PR #877, opened 2024-06-28
Use sqlite for bundle

namse:
Use sqlite for bundle will make a big performance improvement for windows(https://www.sqlite.org/fasterthanfs.html), and also I'd like to change bundle system in wasi more simple with sqlite.

Windows not tested.
Wasi tested.

Breaking changes: no `bundle:` for bundle url.
```rust
// before
let buffer = namui::system::file::bundle::read("bundle:resources/text.txt");

// after
let buffer = namui::system::file::bundle::read("resources/text.txt");
```


namse:
@bigfoodK Could you make it work on windows?

bigfoodK (approved):

namui/namui-cli/src/services/resource_collect_service.rs
```diff
@@ -149,33 +149,46 @@ impl CollectOperation {
...
-    }
+    let dest_dir_path = dest_dir_path.as_ref();
+    debug_println!(
+        "resource_collect_service: copy {:?} -> {:?}/{:?}",
+        &from,
+        &dest_dir_path,
+        &from.file_name().unwrap()
+    );
+    ensure_dir(dest_dir_path)?;
+    assert!(dest_dir_path.is_file());
```
bigfoodK, line 174: You mean `assert!(dest_dir_path.is_dir());`?

namse: oh, `assert!(from.is_file())`;

namse:
```suggestion
    assert!(from.is_file());
```

PR #898, opened 2024-07-19
Support ToOnwed for Dependencies

namse:
Previous version didn't support `&str -> String` so I added that support


namse:
Shit, I didn't know that Dependencies must be sig to track changes

namse:
Continues on #899

PR #908, opened 2024-08-07
Support map for Rect

namse:


namui/namui-type/src/types/units/rect.rs
```diff
@@ -625,3 +625,40 @@ where
         }
     }
 }
+
+impl<T> Rect<T>
+where
+    T: Debug + rkyv::Archive,
+    <T as rkyv::Archive>::Archived: Debug,
```
bigfoodK, line 632: `T: Debug + rkyv::Archive` is required for map?

namse: that;s very right I thought it is required

bigfoodK (approved):

PR #922, opened 2024-08-28
Setup rayon concurrency on namui init

bigfoodK:


namui/namui-cli/webCode/src/imports/hardwareConcurrency .ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export function hardwareConcurrencyHandleOnMainThread(): {
+    onHardwareConcurrency: (
+        payload: WorkerMessagePayload & { type: "hardware-concurrency" },
+    ) => void;
+} {
+    return {
+        onHardwareConcurrency({ concurrencyBuffer }) {
+            const concurrency = navigator.hardwareConcurrency;
```
namse, line 41: Is it only accessible from main thread?

bigfoodK: I thought so, but it wasn't.
It is also accessible from the navigator of worker thread
Let me fix it

namui/namui/src/namui/system/wasi/mod.rs
```diff
@@ -71,6 +71,12 @@ impl Drop for JsHandle {
     }
 }
 
+pub(crate) async fn hardware_concurrency() -> u32 {
+    tokio::task::spawn_blocking(|| unsafe { _hardware_concurrency() })
```
namse, line 75: sync please

PR #932, opened 2024-08-31
Namui New http wrapping

namse:
This PR exposes [http](https://docs.rs/http/latest/http/) crate as interface for http request and response.
I put some Extension traits of http::Request and http::Response, so user can send and receive using that trait.

# Implementation Status

Windows: need tokio update (https://github.com/tokio-rs/tokio/pull/6822), because hyper-util's dns system use original tokio.
Wasi: Works well.
image

google's fail is natural because google server blocks our request by cors policy.

# Additional changes

1. `new-event-system` added. I will put every js-rs event things in here, including pre-made other event systems, like websocket, screen, etc.

### DIFF
--- a/namui/namui/src/system/wasi/new_event_system/event.rs
+++ b/namui/namui/src/system/wasi/new_event_system/event.rs
@@ -1,0 +1,105 @@
+pub(crate) enum Event {
+    HttpFetchOnResponse {
+        fetch_id: u32,
+        status: u16,
+        headers: Vec<(String, String)>,
+    },
+    HttpFetchOnResponseBodyChunk {
+        fetch_id: u32,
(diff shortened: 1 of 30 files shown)


namui/namui/src/system/wasi/new_event_system/event.rs
```diff
@@ -0,0 +1,105 @@
...
+        fetch_id: u32,
+    },
+    HttpFetchOnError {
+        fetch_id: u32,
+        message: String,
+    },
+    BufferPoolRequestBuffer,
+}
+
+// export type NewSystemEvent =
```
namse, line 22: This is for Copilot's inference. Let me keep this.

bigfoodK (approved):

PR #941, opened 2024-09-08
Impl RequestExt for Request<Vec<u8>>

bigfoodK:
Only Impl RequestExt for Rquest<()> existed. So I couldn't send bytes as a request body.


namse:
Already implemented in https://github.com/NamseEnt/namseent/pull/939

PR #946, opened 2024-09-10
Add asset mange page to upload asset

bigfoodK:
Asset upload feature only

### DIFF
--- a/luda-editor/new-client/src/asset_manage_page.rs
+++ b/luda-editor/new-client/src/asset_manage_page.rs
@@ -96,7 +96,7 @@ async fn select_asset_file() -> Result<SelectedAssetFile> {
     return Ok(SelectedAssetFile { name, bytes });
 
     async fn try_read_i32(rx: &mut UnboundedReceiver<Vec<u8>>) -> Result<i32> {
-        let bytes = rx.recv().await.unwrap();
+        let bytes = rx.recv().await.ok_or(anyhow!("data channel closed"))?;
         if bytes.len() != 4 {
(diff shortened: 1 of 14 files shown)


luda-editor/new-client/src/select_asset_file.js
```diff
@@ -0,0 +1,69 @@
...
+        if (result.done) {
+            namui_sendData(encodeNumberToInt32Bytes(0));
+            break;
+        }
+
+        const buffer = result.value.buffer;
+        for (let offset = 0; offset < buffer.byteLength; offset += CHUNK_SIZE) {
+            const chunk = buffer.slice(offset, offset + CHUNK_SIZE);
+            namui_sendData(encodeNumberToInt32Bytes(chunk.byteLength));
+            namui_sendData(chunk);
```
namse, line 58: why did you send data in chunked way?

namse: Use https://github.com/NamseEnt/namseent/pull/948 and you don't need share length of str or bytes because data sending always keep the order.

luda-editor/new-client/src/asset_manage_page.rs
```diff
@@ -0,0 +1,165 @@
...
+                    "invalid chunk bytes length: {}, {} expected",
+                    chunk.len(),
+                    chunk_length
+                ));
+            }
+            read_count += chunk_length;
+            writer.write_all(&chunk)?;
+        }
+        Ok(file_bytes)
+    }
```
namse, line 150: You don't need this complicated code. just send data from js twice and translate it to string and vec.

namse (approved):
LGTM

PR #949, opened 2024-09-11
Support audio, remove media

namse:
The reason why I remove media is, I think we don't need to embed ffmpeg, that's too big to carry to wasm.
And most of case for us, the codecs are very specific, like, opus for audio and video for av1. that mean we don't need ffmpeg.

So I removed media, and add audio system.

## Not implemented in this PR

- Stop
- Resume
- Pause

## Environment changes

- Linux
  - need to install alsa. for ubuntu, `sudo apt-get install -y libasound2-dev`
- Linux host, msvc build
  - need to install ninja. `sudo apt-get install -y ninja-build`

Tested with media sample, in wasi and windows.


namui/namui/src/system/audio/cpal.rs
```diff
@@ -0,0 +1,537 @@
+use super::InterleavedAllSamples;
```
namse, line 1: This file is big so github collapsed this. please check this file too.

namse:
This PR requires the header file typo issue in Opus to be merged first. I've submitted a PR for it: https://github.com/DoumanAsh/opusic-sys/pull/13

PR #953, opened 2024-09-13
Add doc name in primary key

namse:
Previously, the primary key was the composite key (pk, sk), so If pk and sk are same with other doc, it overwrite that.

`TeamNameToTeamIdDoc` was the same with `TeamDoc` so I put name also in composite primary key like (name, pk, sk).

several changes to help debug.


namse:
Review and merge https://github.com/NamseEnt/namseent/pull/952 first please.

namse:
You must have to run `./clear_storage` and `./init_storage` after merge this pr.

bigfoodK (approved):

PR #969, opened 2024-09-25
Put EpisodeEditingUserDoc on creating episode

bigfoodK:
'EpisodeNotExist' occurs because `EpisodeEditingUserDoc` is not created

https://github.com/NamseEnt/namseent/blob/0ab5b050d36b6d27ea1da2e6c6b5c4c9bddf4551/luda-editor/new-server/server/src/api/episode_editor/join_episode_editor/mod.rs#L80-L117


namse (approved):
Good

PR #974, opened 2024-10-14
Add go back buttons

bigfoodK:
image
image
image
image
image


namse:
Conflict resolve please~ @bigfoodK

namse (approved):
I guess you need to make a stack on route so back button just pop it up.

PR #978, opened 2024-10-22
Fix table fit and add manual system init for test

namse:


namse:
ignore wine install error on CI. Will fixed later naturally.

bigfoodK (changes requested):

bigfoodK (approved):

PR #985, opened 2024-11-25
Add player

bigfoodK:
- There is psd sprite clipping issue.  #987

image
image


bigfoodK:
Currently, the player is implemented outside of the editor, so every time you open the player, you leave the editor.  It needs to be re-implemented to see the preview inside the editor.

PR #1023, opened 2025-02-11
Apply status effects

namse:


tower-defense/src/game_state/tower/mod.rs
```diff
@@ -70,11 +101,37 @@ impl Deref for Tower {
...
+        if tower.cooldown == Duration::from_secs(0) {
+            return;
+        }
+
+        let mut cooldown_sub = dt;
+        tower.status_effects.iter().for_each(|status_effect| {
+            if let TowerStatusEffectKind::AttackSpeedAdd { add } = status_effect.kind {
+                cooldown_sub += Duration::from_secs_f32(add);
+            }
+        });
```
namse, line 124: This is problematic. `add` is not second.

bigfoodK: How about using the following methods?

- On shoot, store shoot time and calculate next shoot time
- The next shoot time can be calculated as shoot time + (1 sec/attack speed)
   - At this point, the attack speed means how many times per second tower can attack
- If the attack speed buff is applied after launch, just recalculate the next shoot time

How do you think?

namse: I tried new way to calculate it. please check it again.

bigfoodK (approved):

PR #1025, opened 2025-02-12
Generate items randomly

bigfoodK:


tower-defense/src/game_state/item.rs
```diff
@@ -96,3 +101,303 @@ impl Item {
...
+            }
+        }
+        ItemCandidate::TowerSpeedPlus => {
+            let mut rng = thread_rng();
+            let amount = match rarity {
+                Rarity::Common => rng.gen_range(0.1..0.25),
+                Rarity::Rare => rng.gen_range(0.25..0.5),
+                Rarity::Epic => rng.gen_range(0.5..0.75),
+                Rarity::Legendary => rng.gen_range(0.75..1.0),
+            };
```
namse, line 204: I suggest this code to be

```rust
let amount = thread_rng().gen_range(match rarity { Rarity::Common => 0.1..0.25, .. })
```

This must be more readable.

namse (changes requested):

PR #1040, opened 2025-03-05
Fix tower flickering

bigfoodK:


tower-defense/src/asset_loader.rs
```diff
@@ -0,0 +1,101 @@
...
+                                return;
+                            };
+                            set_tower_asset_loader.mutate(move |tower_asset_loader| {
+                                tower_asset_loader.set_asset(tower_kind, animation_kind, image);
+                            });
+                        }
+                    })
+                });
+
+                join_all(futures).await;
```
namse, line 54: Actually you don't have to join all if you don't handle the result. Instead, spawn the future in runtime is enough. I guess you will check futures failure

namse (approved):

tower-defense/src/asset_loader.rs
bigfoodK: Really? Then I'll just remove it

tower-defense/src/asset_loader.rs
```diff
@@ -0,0 +1,99 @@
...
+                                load_image_from_resource_location(resource_location).await
+                            else {
+                                return;
+                            };
+                            set_tower_asset_loader.mutate(move |tower_asset_loader| {
+                                tower_asset_loader.set_asset(tower_kind, animation_kind, image);
+                            });
+                        }
+                    })
+                });
```
namse, line 52: No, This won't work. You should pass future to runtime's spawn fn as paramter

bigfoodK: Do you mean we should use `ctx.spawn` instead of `namui::spawn`?

namse (approved):

PR #1050, opened 2025-03-28
Give quest reward

bigfoodK:


tower-defense/src/game_state/quest/tracking.rs
```diff
@@ -373,12 +370,6 @@ pub fn on_quest_trigger_event(game_state: &mut GameState, event: QuestTriggerEve
                         });
                     }
                 }
-                QuestTriggerEvent::Reroll => {
-                    remove_quests.push(RemoveQuest {
-                        index: quest_index,
-                        completed: false,
-                    });
-                }
```
namse, line 381: Why did you removed this and UseItem Event?

bigfoodK: I don't want to consider failing quests now

namse (approved):

PR #1068, opened 2025-04-09
Show tower select upgraded value on preview

namse:


tower-defense/src/tower_selecting_hand/mod.rs
```diff
@@ -104,6 +103,7 @@ impl Component for TowerSelectingHand<'_> {
                                     ctx.add(TowerPreview {
                                         wh,
                                         tower_template: &tower_template,
+                                        selected_card_count: using_cards.len(),
```
bigfoodK, line 106: `selected_card_count` can be a problem.
Currently, we choose the highest combination that user can make with the selected cards.

For example, if a one-pair tower is to be built with all the cards selected, there are actually two cards used, but the selected_card_count is 5.

Instead of `selected_card_count`, we should use `tower_kind`, or we should implement additional logic to count the actual number of cards used to build the tower.

bigfoodK (changes requested):

bigfoodK (approved):

PR #1069, opened 2025-04-09
Fix left reroll chance counting for prevent error from extra reroll

bigfoodK:
It was the best record, but the game crashed because of this

### DIFF
--- a/tower-defense/src/tower_selecting_hand/mod.rs
+++ b/tower-defense/src/tower_selecting_hand/mod.rs
@@ -68,6 +68,7 @@ impl Component for TowerSelectingHand<'_> {
                     cards[index] = Card::new_random();
                 }
                 game_state.left_reroll_chance -= 1;
+                game_state.rerolled_count += 1;
 
                 on_quest_trigger_event(game_state, QuestTriggerEvent::Reroll);
             });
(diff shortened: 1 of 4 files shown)


tower-defense/src/tower_selecting_hand/mod.rs
```diff
@@ -68,6 +68,7 @@ impl Component for TowerSelectingHand<'_> {
                     cards[index] = Card::new_random();
                 }
                 game_state.left_reroll_chance -= 1;
+                game_state.rerolled_count += 1;
```
namse, line 71: I think it's not a good idea to keep two opposite data in game_state.

namse (changes requested):

tower-defense/src/tower_selecting_hand/mod.rs
bigfoodK: Can you suggest a better way?

namse: Keep only one of them. Why you need both field?

bigfoodK: https://github.com/NamseEnt/namseent/pull/1069/files#diff-438a654ec124f4f54287542e6146773dc7ac909ba3ed03d6f8bd7acf61a73caaL116-L118

An error occurred in the code above because the `extraroll item` does 'left_reroll_chance += 1'.

namse: then I guess `extraroll item` should affect to max reroll count.

namse: (max reroll count + rerolled count) vs (rerolled count + left reroll chance)
Ok, both are 2 field. OK

PR #1081, opened 2025-04-23
Use global xy for mouse xy in test

namse:
I found that mouse xy in test doesn't work well with ``ctx.translate`.
As quick fix, I removed converting mouse xy to local xy during bounding box check.

It mean, if we rotate Rect and test mouse in, it will return false answer on the left-top, right top, and bottoms also because it only handle with bounding box, not path itself.

I will change that later.


namse:
Something is wrong. Test Failed in comike application

bigfoodK (approved):

PR #1083, opened 2025-05-01
Component support for &Option<T>

namse:


namui/namui-hooks/src/component/component_trait.rs
```diff
@@ -35,11 +27,20 @@ impl Component for RenderingTree {
...
 impl<T: Component> Component for Option<T> {
     fn render(self, ctx: &RenderCtx) {
-        ctx.compose(|ctx| {
-            if let Some(v) = self {
-                ctx.add(v);
-            }
-        });
+        if let Some(v) = self {
+            ctx.add(v);
+        }
```
bigfoodK, line 32: Is it safe to selectively render the v without compose even when the v has a state?

namse: Very safe

bigfoodK (approved):

PR #1141, opened 2025-08-06
Fix wrong tower placement test

namse:
I found that tower placement test's optimization is wrong, so I removed that. I expect it wouldn't affect speed of function.


tower-defense/src/game_state/can_place_tower.rs
```diff
@@ -10,17 +8,6 @@ pub fn can_place_tower(
     route_coords: &[MapCoord],
     map_wh: Wh<usize>,
 ) -> bool {
-    // NOTE: I guess this function would be called frequently so it needs to be optimized.
```
namse, line 13: main changes

namse:
Will merge after #1137 merged

namse:
No, it must be merged before 1137 becuase 1137 needs changes of this PR.

PR #1165, opened 2025-08-17
Prevent panic on field particle system

bigfoodK:


namse:
No, 'initial vec' must have content. unwraping is right.

bigfoodK:
> Let's use ctx.add(fn(&RenderCtx)₩ for on render function. that will fix compose key problem without implicit key.

You mean here?
https://github.com/NamseEnt/namseent/pull/1165/files#diff-f3ce860059c3fc8b06e9cef8f3341200b50a569ae76702f526fa64a76a7fed64R4-R14

namse:
#1170 will solve that problem.

PR #1201, opened 2025-10-31
Fix tower overlaped construction

namse:
https://github.com/user-attachments/assets/96b27f6d-4418-47f6-b1a6-f905b04f0b90


namse:
Yes. lock files are included always

PR #1232, opened 2026-02-05
Support atlas

bigfoodK:
Not fully tested. It's been experimentally introduced


bigfoodK:
for optimize with high particle counts. is there other way?

PR #1234, opened 2026-02-09
Compact time types to f32 4byte

namse:


bigfoodK (approved):

namse:
Tested on local tower defense

PR #1272, opened 2026-03-26
Apply paper texture to icon

bigfoodK:
image


namse:
Hmm I guess it's not a good idea. Because it looks same layer both popup and icons and background by seamless pattern. I suggest not to do this

PR #1275, opened 2026-04-05
Rework upgrade system

bigfoodK:
- split tower upgrades and treasure
- user can use shop every run
- user can buy item and tower upgrade with gold
- user can choose treasure after boss run
- use full card deck and add eraser treasure that removes some card from deck

image


namse (approved):
Good mood!

PR #1318, opened 2026-06-12
Minor ux improvement

bigfoodK:
- Move camera with mouse left, middle right buttons
- Reroll all cards if no selected
- Remove eraser upgrade


namse (approved):
I'm not sure about this

```
if (button == MouseButton::Left)
                            | (button == MouseButton::Middle)
                            | (button == MouseButton::Right)
                        {
                            set_middle_mouse_button_dragging.set(None);
                        }
```

PR #1330, opened 2026-06-17
Add words and use stacked tooltip system

bigfoodK:
- Show tooltip for topbar values
- Use word based description and show related word tooltip

image
image


namse (approved):
I guess you should highlight `keywords` with other color, not the same white color with normal text.

